PR #22, opened 2019-10-14
18 scenario as object

RomainGoussault:


data_splitting.py
```diff
@@ -14,38 +14,41 @@
...
+    x_train = scenario.x_train
+    y_train = scenario.y_train
+    x_test = scenario.x_test
+    y_test = scenario.y_test
+    
     # Describe data
-    print('\n### Data loaded: ', dataset_name)
-    print('- ' + str(len(x_train)) + ' train data with ' + str(len(y_train)) + ' labels')
+    print('\n### Data loaded: ', scenario.dataset_name)
+    print('- ' + str(len(scenario.x_train)) + ' train data with ' + str(len(y_train)) + ' labels')
```
bowni: `y_train` should be prefixed with `scenario.` too

data_splitting.py
```diff
@@ -114,21 +117,20 @@ def process_data_splitting_scenario(x_train, y_train, x_test, y_test, dataset_na
...
             y_node_test = y_test[test_idx]
-        elif testset_option == 'Global':
+        elif scenario.testset_option == 'Global':
             x_node_test = x_test
             y_node_test = y_test
         else:
-            print('This testset_option scenario is not recognized. Quitting with quit()')
-            quit()
-        
+            raise NameError('This testset_option scenario is not recognized. Quitting with quit()')
```
bowni: You should remove "Quitting with quit()" for coherence here :)

contributivity.py
```diff
@@ -0,0 +1,23 @@
...
+class Contributivity:
+  def __init__(self, name='', contributivity_scores=[], computation_time=0):
+    self.name = name
+    self.contributivity_scores = contributivity_scores
+    self.computation_time = computation_time
+    
+    
+  def __str__(self):
+    output = self.name + '\n'
+    output += 'computation time: ' + str(self.computation_time) + ' s' + '\n'
```
bowni: Looks like the whitespace before the "s" of "seconds" is a typo - it would be prettier without.

data_splitting.py
RomainGoussault: will do

data_splitting.py
RomainGoussault: will do

contributivity.py
RomainGoussault: Hum no it's not a typo :)  --> https://english.stackexchange.com/questions/2794/punctuation-with-units

PR #31, opened 2019-11-25
Develop-celine

celinejacques:


data_splitting.py
```diff
@@ -121,8 +122,10 @@ def process_data_splitting_scenario(scenario):
         else:
             raise NameError('This testset_option [' + scenario.testset_option + '] scenario is not recognized')
             
-        node = Node(x_node_train, x_node_test, y_node_train, y_node_test)
+        node = Node(x_node_train, x_node_test, y_node_train, y_node_test, 
+                    str(nodeId))
```
bowni: Are you sure @celinejacques that it has to be stringified here? Looks like in the rest of the code you keep it an integer and stringify it each time you print it.

data_splitting.py
```diff
@@ -105,6 +105,7 @@ def process_data_splitting_scenario(scenario):
   
     # Populate nodes
     node_list = []
+    nodeId=1
```
bowni: @celinejacques could you reformat it to `node_id` to enforce the snake_case style?

fl_training.py
```diff
@@ -138,8 +138,8 @@ def compute_test_score(node_list):
                           verbose=0,
                           validation_data=(node.x_val, node.y_val))
                 
-                val_acc_list.append(history.history['val_acc'])
-                acc_list.append(history.history['acc'])
+                val_acc_list.append(history.history['val_accuracy'])
+                acc_list.append(history.history['accuracy'])
```
bowni: For my culture, is this a Keras usage best practice?

data_splitting.py
celinejacques: My idea was that the name will be a string such as the name of the center. 
Which option do you prefer: 
- I change node number to string before initialize it (so nodeId or node_id now is a string and not stringify it here)
- I left it as a number and stringify it only for print. 
- Other idea? 
I will change the prints when it is needed with the option you prefer.

fl_training.py
celinejacques: Sorry in the first version that I used to test 'acc' did not work. I needed to write 'accuracy' and 'val_accuracy' (same for Samya). But I had to install again everything and now it works with 'acc'.

bowni:
Thanks @celinejacques , pretty cool! @RomainGoussault I let you merge? As you are currently developing on a number of issues, I think it is best that you merge yourself so that you can then locally pull and rebase your current work.

data_splitting.py
```diff
@@ -105,6 +105,7 @@ def process_data_splitting_scenario(scenario):
   
     # Populate nodes
     node_list = []
+    node_id=1
```
RomainGoussault: We should start at 0.

celinejacques: @RomainGoussault No problem, it is done

RomainGoussault:
Thanks @celinejacques !

Issue #33, opened 2019-11-26
Add performance in result file


RomainGoussault:
Add performance in result file, this will give an indication of how good was the (federated) training.

RomainGoussault:
done

Issue #23, opened 2019-10-16
Add visualisation for training


RomainGoussault:
Ideally with TensorBoard ?

RomainGoussault:
Visualisation is composed of :

- data visualisation for each node
- training on every node
- global model training

RomainGoussault:
done in PR #34

Issue #3, opened 2019-10-04
Implement epochs EarlyStopping


bowni:
Implement epochs [EarlyStopping](https://keras.io/callbacks/#earlystopping) to optimize computational cost of simulations.

bowni:
@RomainGoussault this issue could go in a next milestone (0.2), I don't think it is needed to present and demo the project.

RomainGoussault:
Done in PR #34

PR #34, opened 2019-11-27
Dev romain 

RomainGoussault:


RomainGoussault:
Fix issues #33 #23 #3

fl_training.py
```diff
@@ -114,7 +123,32 @@ def compute_test_score(node_list):
...
+                                                             verbose=0)
+                current_val_loss = model_evaluation[0]
+                global_val_acc.append(model_evaluation[1])
+                global_val_loss.append(current_val_loss)
+
+                # Early stopping
+                if is_early_stopping:
+                    
+                    # Early stopping parameters
+                    patience = 4
```
bowni: @RomainGoussault shouldn't we put this `patience` parameter in `constants.py`?

scenario.py
```diff
@@ -65,21 +219,18 @@ def to_file(self):
     out += 'Percentages of data samples per node: ' + str(self.amounts_per_node) + '\n'
     out += 'Random or stratified split of data samples: ' + self.samples_split_option + '\n'
     out += 'Centralised or distributed test set: ' + self.testset_option + '\n'
-    out += 'Number of epochs defined in learning algos: ' + str(self.nb_epochs) + '\n'
+    out += 'Number of epochs: ' + str(self.epoch_count) + '\n'
+    out += 'Early stopping activé: ' + str(self.is_early_stopping) + '\n'
```
bowni:
```suggestion
    out += 'Early stopping on? ' + str(self.is_early_stopping) + '\n'
```

simulation_run.py
```diff
@@ -14,32 +14,23 @@
...
-
 # Create a custom scenario and comment the main scenario parameters (see scenario.py for more comments)
-my_custom_scenario = scenario.Scenario()
+my_custom_scenario = scenario.Scenario(is_quick_demo=True)
 my_custom_scenario.nodes_count = 3 # Number of nodes in the collaborative ML project simulated
 my_custom_scenario.amounts_per_node = [0.20, 0.30, 0.5] # Percentages of the data samples for each node
-my_custom_scenario.samples_split_option = 'Stratified' # If data are split randomly between nodes or stratified to be distinct (toggle between 'Random' and 'Stratified')
+my_custom_scenario.samples_split_option = 'Random' # If data are split randomly between nodes or stratified to be distinct (toggle between 'Random' and 'Stratified')
 my_custom_scenario.testset_option = 'Centralised' # If test data are distributed between nodes or stays a central testset (toggle between 'Centralised' and 'Distributed')
-my_custom_scenario.nb_epochs = constants.NB_EPOCHS
```
bowni: I suggest to leave the line `my_custom_scenario.epoch_count = 15` here so that the user knows easily what can be configured in a custom scenario @RomainGoussault

scenario.py
```diff
@@ -42,19 +46,169 @@ def __init__(self):
...
+        self.y_test = y_test[:50]
+        
+        self.epoch_count = 2
+        
+    
   def append_contributivity(self, contributivity):
             
       self.contributivity_list.append(contributivity)
+          
+  def split_data(self):
```
bowni: I suppose it makes a lot more sense to have it here, thanks!

simulation_run.py
RomainGoussault: If we leave it here, the boolean `is_quick_demo` in the Scenario constructor will not work anymore. 

If we want to leave `my_custom_scenario.nb_epochs`, we wil have to re-work the `is_quick_demo` part

PR #37, opened 2019-11-29
Truncated monte carlo

Thomas-Galtier:


bowni:
Hi @Thomas-Galtier , looks super promising I am eager to test this.

Just a suggestion for the future, to ensure lisibility in the global commit history: I recommend following those best practices in the formulation of commit messages: [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/). It's really good (IMO).

truncated_MonteCarlo.py
```diff
@@ -0,0 +1,96 @@
...
+            t+=1
+            print(t<50,t<=norm.ppf(0.975, loc=0, scale=1)**2* np.max(mean_squares_sv-sv**2) /sv_accuracy**2,firsts_are_equal)
+            print(t)
+            print(sv)
+            previous_sv = np.copy(sv)#store previous estimation of the shapley values
+            previous_mean_square_sv= np.copy(mean_squares_sv) #store previous value of   the mean_square_sv
+            previous_permutation=np.copy(permutation)# store the previous permutation
+            permutation = np.random.permutation(n) # store the current permutation
+            if firsts_are_equal and t>1:# makes sure we don't stop the loop because the first permutations are the same (very unlikely but can still happen if n is small)
+                firsts_are_equal=   (permutation==previous_permutation).all()
```
bowni: The `while` loop has an `or` conditions, so this variable is not going to be used before `t` reaches `50`. I think there is something I don't understand, sorry.

truncated_MonteCarlo.py
```diff
@@ -0,0 +1,96 @@
...
+    if n==1:
+        return np.array([1])
+    else:
+        characteristic_all_node= not_twice_characteristic(np.arange(n)) #characteristic function on all nodes
+        sv = np.zeros(n) #store current estimation of the shapley values 
+        mean_squares_sv = np.zeros(n) # used for variance computation in the condition of the while loop below 
+        permutation=np.zeros(n) # store the current permutation 
+        firsts_are_equal =True
+        t=0
+        while t<50 or t<=norm.ppf(0.975, loc=0, scale=1)**2* np.max(mean_squares_sv-sv**2) /sv_accuracy**2 or firsts_are_equal: #check if the length of the confidence interval  is below the value of sv_accuracy
```
bowni: Shouldn't it be an `and` condition?

bowni: In `t<=norm.ppf(0.975, loc=0, scale=1)**2* np.max(mean_squares_sv-sv**2) /sv_accuracy**2` I don't understand what this quantity represents and why/how you relate it to `t`. Could you please explain?

truncated_MonteCarlo.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def truncated_MC(preprocessed_node_list, sv_accuracy=0.001, contrib_accuracy=0.01):
+    """Return the vector of approximated shapeley value corresponding to a list of node and a characteristic function using the truncated monte-carlo method."""
+    
+    characteristic_func=fl_training.compute_test_score
+    n = len(preprocessed_node_list)
+    
+    # we store the value of the characteristic function in order to avoid recomputing it twice
+    char_value_dict={():0} # the dictionary that will countain the values
```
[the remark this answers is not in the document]
Thomas-Galtier: Yes that is it

truncated_MonteCarlo.py
```diff
@@ -0,0 +1,96 @@
...
+    listn=np.arange(n)
+    
+    # return the characteristic function of the nodelist associated to the ensemble  permut, without recomputing it if it was already computed
+    def not_twice_characteristic(permut):
+        #sort permut
```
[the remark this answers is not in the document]
Thomas-Galtier: Here the list we want to sort is a array of indices so I think it goes faster this way, but i am not 100% sure.

truncated_MonteCarlo.py
Thomas-Galtier: No it is well an `or` condition. For two reasons: 
1) The approximations of the Shapeley values relies on the Central limit theorem which is an asymptotic result. We want to have at least `50` iterations so that we can say we are in an asymptotic situation, and that our approximations are located around the true Shapeley values.
2) The quantity `mean_squares_sv-sv**2` is supposed to asymptotically approximate the variance of the contributions so `t` needs to be at least 50 so that this approximation (and therefore the second condition) is meaningfull.

Thomas-Galtier: For the second condition I think we should change it because it it too strong...  Nevertheless the idea behind this condition is to make sure that half the length of of the confidence interval for each approximation of a Shapeley value is lower than `sv_accuarcy`, so that the approximation errors are below `sv_accuarcy ` with 95% chance.

let`q=norm.ppf(0.975, loc=0, scale=1)` and `v=mean_squares_sv-sv**2`. Let `L[i]` be half the length of the confidence interval for the i-th Shapeley value approximation then `L[i]=q*np.sqrt(v[i]/t)`. The condition `t<=norm.ppf(0.975, loc=0, scale=1)**2* np.max(mean_squares_sv-sv**2) /sv_accuracy**2` is equivalent to `np.max(L)<sv_accuracy`.

Thomas-Galtier: This condition impose a very high number of iterations so I think we should change it. We should probably replace tit by something like `np.max(np.mean(contributions,axis=0) - np.mean(contributions[0:t-1],axis=0))>sv_accuracy`

truncated_MonteCarlo.py
Thomas-Galtier: Yes but assume we have only the two first conditions in the while. If the 50 first permutations drawn are the same then np.max(mean_squares_sv-sv**2)=0 and the second condition makes the loop stop. firsts_are_equal is here to avoid this early stopping case.

truncated_MonteCarlo.py
Thomas-Galtier: I finally kept the original condition.

truncated_MonteCarlo.py
bowni: Oh, understood. I didn't even consider the possibility of the first 50 being the same :)

bowni: Oh, understood. I didn't even consider the possibility of the first 50 being the same :)

bowni:
Hmm, weird, this had never happened to me before: the merge (via Github GUI) worked, this can be viewed in `master` git history, but on Github GUI the PR is still open and now says can't be merged because of conflicts... Let's see in a couple of hours/days.

Issue #35, opened 2019-11-27
Review model evaluation


RomainGoussault:
@bowni  this looks weird:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/master/fl_training.py#L166

We always evaluate the model on the last node

RomainGoussault:
One more thing to do : correct early stopping so it uses validation data and not test data. Actually validation data should not be used per node but only per scenario to do early stopping.

bowni:
See PR #43

PR #43, opened 2020-01-13
35 review model evaluation

bowni:
Corresponds to issue #35

### DIFF
--- a/fl_training.py
+++ b/fl_training.py
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
 """
-Created on Fri Aug 30 15:33:15 2019
-
-@author: RGOUSSAULT
-inspired from: https://keras.io/examples/mnist_cnn/
+Train a model across multiple nodes
+(inspired from: https://keras.io/examples/mnist_cnn/)
 """
 
 from __future__ import print_function
@@ -73,7 +71,7 @@ def compute_test_score_for_single_node(node, epoch_count):
     # Return model score on test data
     return model_eval_score
 
-# TODO no methods overloadin
+#%% TODO no methods overloading
 def compute_test_score_with_scenario(scenario, is_save_fig=False):
     return compute_test_score(scenario.node_list, 
                               scenario.epoch_count, 
@@ -82,7 +80,7 @@ def compute_test_score_with_scenario(scenario, is_save_fig=False):
                               save_folder=scenario.save_folder)
         
         
-# Distributed learning training      
+#%% Distributed learning training      
 def compute_test_score(node_list, epoch_count, is_early_stopping=True, is_save_fig=False, save_folder=''):
     """Return the score on test data of a final aggregated model trained in a federated way on each node"""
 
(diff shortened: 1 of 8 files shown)


scenario.py
```diff
@@ -22,8 +18,8 @@ def __init__(self, is_quick_demo=False):
       
     self.dataset_name = 'MNIST'
     (x_train, y_train), (x_test, y_test) = mnist.load_data()    
-    self.x_train = x_train
-    self.y_train = y_train
+    # The train set has to be split into a train set and a validation set for early stopping (called 'esval' below)
+    self.x_train, self.x_esval, self.y_train, self.y_esval = train_test_split(x_train, y_train, test_size=0.2, random_state=42)
```
RomainGoussault: Validation sets are usually referred as `x_val `and `y_val` not `..._esval`

fl_training.py
```diff
@@ -38,88 +36,110 @@ def preprocess_node_list(node_list):
...
     nodes_count = len(node_list)
-        
+    
+    # If only one node, fall back to dedicated single node function    
     if nodes_count == 1:
         return compute_test_score_for_single_node(node_list[0], epoch_count)
     
+    # Else, follow a federated learning procedure
     else:
+        print('\n## Training and evaluating model on multiple nodes: ' + str(node_list))
```
RomainGoussault: This is the output on my machine:
> ## Training and evaluating model on multiple nodes: [<node.Node object at 0x000001F347E2B748>, <node.Node object at 0x000001F347DCEF98>]
Because Node class does not have` __str__` or `__repr__` methods implemented --> https://stackoverflow.com/questions/4932438/how-to-create-a-custom-string-representation-for-a-class-object

fl_training.py
```diff
@@ -38,88 +36,110 @@ def preprocess_node_list(node_list):
...
     """Return the score on test data of a model trained on a single node"""
     
+    print('\n## Training and evaluating model on one single node.')
     # Initialize model
     model = utils.generate_new_cnn_model()
     # print(model.summary())
 
     # Train model
-    print('\n### Training model on one single node: node ' + node.node_id)
+    print('\n### Training model on one single node: ' + str(node))
```
RomainGoussault: This is the output on my machine:
> ### Training model on one single node: <node.Node object at 0x000001F23B351F60>
Because Node class does not have` __str__` or `__repr__` methods implemented --> https://stackoverflow.com/questions/4932438/how-to-create-a-custom-string-representation-for-a-class-object

scenario.py
bowni: "esval" was for "Early Stopping validation", and prevent confusion with the `x_val` and `y_val` in the intra-node fitting. Can we keep that way for the moment?

fl_training.py
bowni: Yes I know. I couldn't find an easy way to give a sort of a "name" or an "id" that sticks to a given node, and I figured that printing this memory address was not that bad.

bowni:
@RomainGoussault I commented on your points. I suggest to keep things that way for the moment, let me know if you agree or if you think we should improve those now.

fl_training.py
RomainGoussault: I agree. Then we should not print this line since it does not provide any useful information. Or juste print the length of the node list?

PR #64, opened 2020-03-09
Black format

natct10:
This PR aims at gaining some readability thanks to black formater and to reorder imports. Feel free to ask for edits, I would be happy to discuss about this proposition!


RomainGoussault:
It works on my machine

PR #75, opened 2020-03-26
Refactor pre-processing of data and fix evaluation when training on 1 single node

### LINKED ISSUE #48: Improve perf evaluation in single-partner setting
Currently it evaluates on a test set attached to the node, which is created by the `split_data` method in the `scenario.py` object. Would make more sense to evaluate it on an independent test set so that all single-partner trainings would be evaluated on the same test set.

bowni:
- Fixes #48 
- Refactor of datasetes pre-processing: it was scattered in 5 different places (`simulation_run`, `fl_training`, `node`, `scenario`, `utils`). I think it makes more sense to proceed as follow:
  - The call to pre-process the datasets of the scenario is done in `simulation_run`
  - The function is in `fl_training` as it is quite adherent to machine learning choices. It pre-processes both datastets on each node of the scenario **and** central datasets (testset and early stopping validation set)
  - The function leverages `utils`
  - There is no more pre-processing method in the `node` object
  - In the `scenario` object:
    - the main dataset is loaded and split in 3 central datasets: train + early stopping val + test
    - the `split_data` method is then there to split and distribute the train and test central datasets into local datasets for each node
    - the distributed testsets of each node are then used in `fl_training` in single-partner training cases only if `scenario.testset_option == 'Distributed'`. Otherwise the central testset is used


fl_training.py
```diff
@@ -21,52 +21,44 @@
...
+    for node_index, node in enumerate(scenario.node_list):
+        # Preprocess inputs (x) data
+        node.x_train = utils.preprocess_input(node.x_train)
+        node.x_test = utils.preprocess_input(node.x_test)
+        # Preprocess labels (y) data
+        node.y_train = keras.utils.to_categorical(node.y_train, constants.NUM_CLASSES)
+        node.y_test = keras.utils.to_categorical(node.y_test, constants.NUM_CLASSES)
         # Crete validation dataset
-        x_node_train, x_node_val, y_node_train, y_node_val = train_test_split(
+        node.x_train, node.x_val, node.y_train, node.y_val = train_test_split(
```
RomainGoussault: Note: for now we only use `node.x_val` for plotting figures
However I think we should keep it that way for now and it might be useful when we make our federated learning approcach a bit smarter.

scenario.py
```diff
@@ -20,10 +20,11 @@
 class Scenario:
     def __init__(self, params, experiment_path):
 
+        # Identify and get a dataset for running experiments
         self.dataset_name = "MNIST"
         (x_train, y_train), (x_test, y_test) = mnist.load_data()
-        # The train set has to be split into a train set and a validation set for early stopping (called 'esval' below)
-        self.x_train, self.x_esval, self.y_train, self.y_esval = train_test_split(
+        # The train set has to be split into a train set and a validation set for early stopping
+        self.x_train, self.x_valearlystop, self.y_train, self.y_valearlystop = train_test_split(
```
RomainGoussault: `x_val` and `y_val` are the names traditionnaly used for validation datasets :)

scenario.py
```diff
@@ -53,8 +54,11 @@ def __init__(self, params, experiment_path):
             self.corrupted_nodes = ["not_corrupted"] * self.nodes_count
 
         # Define if test data should be distributed between nodes...
+
+        # ... or if each node should refer to a centralised test set...
+        # ... when training on a single node
         # ... or if each node should refer to a centralised test set
-        self.testset_option = params["testset_option"]
+        self.testset_option = params["testset_option"] # Toggle between 'Centralised' and 'Distributed'
```
RomainGoussault: Does it make this clearer like this : ?

> When training on a single node, the test set can be either the local node test set or the global test set

Then I would rename `testset_option `to `single_test_set_evaluation `= `local `or `global`

fl_training.py
bowni: OK

scenario.py
bowni: Yes, I think we need to have a clear and distinct name for the early stopping validation set, to make sure at the local node level there is no confusion with the actual validation set.

scenario.py
bowni: Ok, changed and clarified in commit ec27f20.

fl_training.py
```diff
@@ -21,52 +21,44 @@
...
             node.x_train, node.y_train, test_size=0.1, random_state=42
```
Thomas-Galtier: Could we add the corruption in the preprocess function?

Issue #15, opened 2019-10-08
Weighted weights averaging ?


RomainGoussault:
If nodes don"t have the same data size :
When aggregating models, we should probably not blindy take the mean of the weights of all nodes but we should take the weighted mean.

RomainGoussault:
Agregating weights can be done in several ways:

- Averaging
- Weighted averaging depending on dataset size
- Weighted averaging depending on incremental performance


https://blog.acolyer.org/2019/08/19/robust-learning-from-untrusted-sources/

bowni:
> Agregating weights can be done in several ways:
> 
> * Averaging
> * Weighted averaging depending on dataset size
> * Weighted averaging depending on incremental performance
> 
> https://blog.acolyer.org/2019/08/19/robust-learning-from-untrusted-sources/

I think we should try to parameterize the way the aggregation is weighted. This could be done maybe with:
- different weighting methods that would return a vector of weights
- a parameter to designate with approach is chosen for a given scenario
- passing this parameter as argument to the FL function
- having the FL function call the appropriate weighting methods to get the weights

What do you think @RomainGoussault @celinejacques @Thomas-Galtier?

bowni:
Done in PR #80

PR #80, opened 2020-03-26
Weight aggregation with data volumes

bowni:
- Fixes issue #15
- Between each `.fit` iteration, the aggregation of weights is now weighted with the data volumes of each node
- For that, `scenario.amounts_per_node` is now passed in `fl_training.compute_test_score`


fl_training.py
```diff
@@ -107,6 +107,7 @@ def compute_test_score_with_scenario(scenario, is_save_fig=False):
         scenario.x_test,
         scenario.y_test,
         scenario.is_early_stopping,
+        scenario.amounts_per_node,
```
RomainGoussault: At this point, I would pass the whole scenario object to the `compute_test_score_with_scenario()` method

fl_training.py
```diff
@@ -161,18 +163,24 @@ def compute_test_score(
                 print("   First epoch, generate model from scratch")
 
             else:
-                # Aggregating phase : averaging the weights
+                # Aggregating phase: averaging the weights
                 print("   Aggregating models weights to build a new model")
-                weights = [model.get_weights() for model in model_list]
+                weights_per_model = [model.get_weights() for model in model_list]
+                weights_per_layer = list(zip(*weights_per_model))
```
RomainGoussault: It is way clearer 👍

fl_training.py
```diff
@@ -161,18 +163,24 @@ def compute_test_score(
...
                 new_weights = list()
 
-                for weights_list_tuple in zip(*weights):  # TODO : make this clearer
-                    new_weights.append(
-                        [
-                            np.array(weights_).mean(axis=0)
-                            for weights_ in zip(*weights_list_tuple)
-                        ]
-                    )
+                # OLD:
```
RomainGoussault: remove commented code

fl_training.py
```diff
@@ -121,6 +122,7 @@ def compute_test_score(
     x_test,
     y_test,
     is_early_stopping=True,
+    aggregation_weights=None,
```
RomainGoussault: I think we should have a param in the scenario class to choose which aggregation we want to perform (for now it will only be: weighted by data size or weighted by nothing)
It will also be easier to test the impact of the weighted average, so we can be sure that weighted average is better than no weight.

fl_training.py
bowni: You mean to `compute_test_score()`?

RomainGoussault: Yes you are right, I meant `compute_test_score()`

bowni: We can't do that simply. Because we need a `compute_test_score()` independent from scenarios, to be used in contributivity measure algorithms when needing to launch federated learning on an arbitrary coalition.

fl_training.py
bowni: Ok, done in 6b0ec01.

fl_training.py
bowni: Ok, done in 1cce0b6.

fl_training.py
RomainGoussault: Ah yes you are right !

RomainGoussault:
I am currently running some scenario to see the performance improvement on heavy imbalanded data distribution between partners.

bowni:
@RomainGoussault I just rebased it with master with the recent merge of the pre-processing refacto and the central datasets renaming.

fl_training.py
```diff
@@ -142,6 +176,13 @@ def compute_test_score(
         global_val_acc = []
         global_val_loss = []
 
+        # Create list of weights for aggregation steps
+        aggregation_weights = []
```
RomainGoussault: I think the `aggregation_weights`  should sum to 1:

- uniform:` [1/nodes_count] * nodes_count`
- data-volume: `[len(node.x_train) for node in node_list] / total_data_size_in_nodes`

et un petit `assert(np.sum(aggregation_weighting ) == 1) `à la fin

RomainGoussault:
First test on 0.95 / 0.05 distribution, there is not much difference. I am trying on 0.99 / 0.01 distribution now.
image

Thomas-Galtier:
Ben on a de meilleurs resultats plus de 50% du temps non?

RomainGoussault:
5 repeats with new data distribution. 
image

fl_training.py
bowni: Ok, doing that now.

bowni: Done in cb625f7.

RomainGoussault:
image
(nothing is like uniform but normalized, it was run before the normalization commit https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/80/commits/cb625f7a9e486189d13c92f4d0c188ff9ee7f032)

PR #83, opened 2020-03-27
Implement distributed learning iterations over mini-batches

### LINKED ISSUE #4: Parameterize size of mini-batches
Parameterize size of mini-batches so that distributed learning and weights aggregating phases can be optimized.

bowni:
Implements and closes #4:
- for each epoch, splits the nodes datasets into mini-batches
- iterates over mini-batches for training individual models and aggregating them


fl_training.py
```diff
@@ -149,191 +194,169 @@ def compute_test_score(
...
                 history = node_model.fit(
-                    node.x_train,
-                    node.y_train,
+                    minibatched_x_train[node_index][minibatch+1],
+                    minibatched_y_train[node_index][minibatch+1],
                     batch_size=constants.BATCH_SIZE,
                     epochs=1,
                     verbose=0,
-                    validation_data=(node.x_val, node.y_val),
+                    validation_data=(minibatched_x_val[node_index][minibatch+1], minibatched_y_val[node_index][minibatch+1]),
```
Thomas-Galtier: I not sure but should we rather keep the hole validation data here?

fl_training.py
```diff
@@ -149,191 +194,169 @@ def compute_test_score(
...
+        minibatched_x_train = [None] * nodes_count
+        minibatched_x_val = [None] * nodes_count
+        minibatched_y_train = [None] * nodes_count
+        minibatched_y_val = [None] * nodes_count
+        split_indices = np.arange(minibatch_count+1) / minibatch_count
+        for node_index, node in enumerate(node_list):
+            minibatched_x_train[node_index] = np.split(node.x_train, (split_indices * len(node.x_train)).astype(int))
+            minibatched_x_val[node_index] = np.split(node.x_val, (split_indices * len(node.x_val)).astype(int))
+            minibatched_y_train[node_index] = np.split(node.y_train, (split_indices * len(node.y_train)).astype(int))
+            minibatched_y_val[node_index] = np.split(node.y_val, (split_indices * len(node.y_val)).astype(int))
```
Thomas-Galtier: I not sure but should we rather keep the whole validation data here?

fl_training.py
bowni: By the whole validation dataset you mean the early stopping validation dataset? If we do that don't we risk some kind of "early stopping overfitting"? cc @RomainGoussault

bowni:
I just rebased with `master` (following recent merge of PR #80) to enable an easier review, test and merge @RomainGoussault.

fl_training.py
Thomas-Galtier: I don't think so, because the validation set is not used during the fit. [see here](https://www.tensorflow.org/api_docs/python/tf/keras/Model )The validation set is put in argument only to track the loss after each epoch.

scenario.py
```diff
@@ -60,11 +60,19 @@ def __init__(self, params, experiment_path):
...
+            self.aggregation_weighting = params["aggregation_weighting"]
+        else:
+            self.aggregation_weighting = "uniform"
+
         self.node_list = []
 
         self.contributivity_list = []
 
         self.epoch_count = 40
+        self.minibatch_count = 20
```
RomainGoussault: Specify value in config file so we can easily launch experiments with different values of minibatch.

RomainGoussault: I would also rename to `minibatch_size `(or even to `federated_learning_batch_size`) as the minibatch_size would generally be greater than the batch_size !

RomainGoussault: I was wondering if we should use ratio or absolute values but I think absolute values is better.
Also we probably need an easy way to specify to not use mini batch at all.

fl_training.py
RomainGoussault: A bit hard to follow.
I propose: `minibatched_x_train[node_index] = np.split(node.x_train, minibatches_count)`

fl_training.py
```diff
@@ -149,191 +194,169 @@ def compute_test_score(
...
+
+        # Iterate over mini-batches for training and aggregation
+        for minibatch in range(minibatch_count):
```
RomainGoussault: `minibatch_index `instead of `minibatch`

fl_training.py
```diff
@@ -149,191 +194,169 @@ def compute_test_score(
...
-                        metrics=["accuracy"],
-                    )
+                print("   Training on node " + str(node_index) + " - " + str(node))
 
-                # Train on whole node local data set
+                # Train on node local data set
                 history = node_model.fit(
-                    node.x_train,
-                    node.y_train,
+                    minibatched_x_train[node_index][minibatch+1],
```
RomainGoussault: Why `+1`?

fl_training.py
RomainGoussault: Good question !!
I think using the whole validation data set would make more sense now, as we only use it for plotting for now

fl_training.py
```diff
@@ -141,6 +143,49 @@ def compute_test_score_with_scenario(scenario, is_save_fig=False):
...
 #%% Distributed learning training
 
+def prepare_aggregation_weights(aggregation_weighting, nodes_count, node_list):
+    """Returns a list of weights for the weighted average aggregation of model weights"""
+
+    aggregation_weights = []
+    if aggregation_weighting == "uniform":
+        aggregation_weights = [1] * nodes_count
+    elif aggregation_weighting == "data-volume":
+        aggregation_weights = [len(node.x_train) for node in node_list]
```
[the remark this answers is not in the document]
Thomas-Galtier: actually the average method normalizes the weights by default so it is not needed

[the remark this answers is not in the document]
bowni: @RomainGoussault when a change is done on another branch and merged on `master`, I think it is cleaner to not re-do it in the new branch you started before that merge. The future merge will handle that.
Although I thought I rebased properly, so now this looks weird. I'll check

fl_training.py
```diff
@@ -149,191 +194,169 @@ def compute_test_score(
     y_val_global,
     x_test,
     y_test,
+    aggregation_weighting="uniform",
+    minibatch_count=20,
```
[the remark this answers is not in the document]
bowni: Yes we do need it here. This method is not always used with a scenario perspective, it can also be used independently to compute the distributed learning on any set of nodes.

scenario.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, params, experiment_path):
         if "corrupted_nodes" in params:
             self.corrupted_nodes = params["corrupted_nodes"]
         else:
-            self.corrupted_nodes = ["not_corrupted"] * self.nodes_count
+            self.corrupted_nodes = ["not-corrupted"] * self.nodes_count
```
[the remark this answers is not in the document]
bowni: Hmmm, this was a way to differentiate variable names (with underscores) and names of variable values (with dashes), of which I am not at the origin. I can change it, but please note it implies changes in the code for corrupting datasets.

scenario.py
bowni: Ok, will do

bowni: Hmm, this I am not a fan. This variable is not for a "size", it is for a "count".

fl_training.py
bowni: I did that in the beginning :-) But it doesn't work, `np.split()` can't accept an integer if it doesn't divide perfectly. So the proper way, as documented in NumPy doc, is to pass indices where to split.

fl_training.py
bowni: Because the first piece of the split is empty, by construction.

contributivity_measures.py
```diff
@@ -53,7 +53,7 @@ def compute_independent_scores(
 #%% Generalization of Shapley Value computation
 
 
-def compute_SV(node_list, epoch_count, x_val_global, y_val_global, x_test, y_test):
+def compute_SV(node_list, epoch_count, x_val_global, y_val_global, x_test, y_test, aggregation_weighting="uniform", minibatch_count=20):
```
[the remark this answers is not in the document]
bowni: Ok done

fl_training.py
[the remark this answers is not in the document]
bowni: Ok, fixed

scenario.py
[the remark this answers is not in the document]
bowni: Ok, done

scenario.py
bowni: Ok, done

fl_training.py
bowni: Ok, done

fl_training.py
bowni: Ok, clarified

fl_training.py
bowni: Ok, done

scenario.py
bowni: I thought about that: I am not sure. Actually, our distributed learning algo doesn't really make sense without the iteration over mini-batches. Iterating only over epochs is really too simplistic, and I don't think we want a parameter to enable the user to do it.

bowni:
To do:

- [ ] Assert that `minibatch_count` < `min(data samples of smaller node)`

fl_training.py
```diff
@@ -150,198 +211,146 @@ def compute_test_score(
...
-                        optimizer="adam",
-                        metrics=["accuracy"],
-                    )
+            # Iterate over nodes for training each individual model
+            for node_index, node in enumerate(node_list):
 
-                # Train on whole node local data set
+                # Train on node local data set
+                print("         Training on node " + str(node_index) + " - " + str(node))
                 history = node_model.fit(
```
RomainGoussault, line 270: `node_model` is not cleared between 2 nodes. It means node_0 trains. Then node_1 trains sequentially on the model trained on node_0. Then node_2 trains sequentially on the model trained on node_0 and node_1 etc.
I think it is valid FL scenario (even probably better than what we did before !) but different to the federated averaging where every node communicates only with a central server/aggregator.

RomainGoussault: In fact we now implemented Sequential FL instead of previous Parallel implementation:
image
(taken from Substra whitepaper)

bowni:
OK @RomainGoussault, I wasn't familiar with this, shame on me!
So I took the following approach: for a given mini-batch, before iterating over all nodes, build a new model for each node. How does that look?

bowni:
Just rebased with recent merge on `master`.

PR #89, opened 2020-04-01
Iss78 contrib method in config

### LINKED ISSUE #78: Add params to choose which contributivity methods to run
* Add params in config file to choose which contributivity methods to run
* Refactor run_scenario() method in simulation_run file so it is not a huge list of the same code pasted X times (start timer, compute contrib, stop timer, add contrib to the list, etc)

celinejacques:
close #78


contributivity_measures.py
```diff
@@ -18,6 +18,48 @@
...
+        # Contributivity 3: Truncated Monte Carlo Shapley
+        tmcs_results = truncated_MC(
+                        current_scenario, sv_accuracy, 
+                        alpha, contrib_accuracy
+                        )
+        score_dict = {"TMCS values": (tmcs_results["sv"], 
+                                      tmcs_results["std_sv"])}
+
+    
+    return score_dict
```
natct10, line 61: Looks good! 
Naive question here: what if no `method_to_compute` is taken by the if/elif conditions (typo or other method name), `score_dict` is returned but not defined?

celinejacques: Good point! I'll add a test when we initiate the scenario and I complete the conditional loop with a `else`.

simulation_run.py
```diff
@@ -101,73 +101,23 @@ def run_scenario(current_scenario):
...
+        for score_method in score_dict:
+            score =  score_dict[score_method]
+            contrib = contributivity.Contributivity(
+                score_method, score[0], score[1], np.round(end - start)
+                )
+    
+            current_scenario.append_contributivity(contrib)
+            print("\n## Evaluating contributivity with " + method + ":")
+            print(contrib)
+
```
RomainGoussault, line 120: C'est clean 😍 !

simulation_run.py
```diff
@@ -101,73 +101,23 @@ def run_scenario(current_scenario):
...
-
+    for method in current_scenario.methods:
+        print(method)
+        start = timer()
+        score_dict = contributivity_measures.compute_contributivity(method, 
+                                                                    current_scenario)
+        end = timer()
+        
+        for score_method in score_dict:
+            score =  score_dict[score_method]
```
RomainGoussault, line 112: double space after `= `

contributivity_measures.py
RomainGoussault: Yes good point to add test when we initiate the scenario so it will fail quickly !

natct10: What you can do here is define `score_dict = {}` at the beginning of the method and remove this `else` condition. So, if a condition is met, `score_dict` is updated otherwise it remains empty and the method return the empty dict.

scenario.py
```diff
@@ -76,13 +77,23 @@ def __init__(self, params, experiment_path):
         else:
             self.epoch_count = 40
             
-        if 'methods' in params:
-            self.methods = params['methods']
+        methods_default = ["Shapley values",
+                           "Independant scores",
+                           "TMCS values"
+                           ]
```
natct10, line 83: For readability, would you kindly format the list as following?
```python
list = [
    "one",
    "two",
    "three",
]

File: contributivity_measures.py (hunk 1 above)
Comment (celinejacques) [reply contributivity_measures.py:62]: Done

File: scenario.py (hunk 4 above)
Comment (celinejacques) [reply scenario.py]: Sure!

File: contributivity_measures.py (hunk 1 above)
Comment (RomainGoussault) [reply contributivity_measures.py:62]: @natct10 I think we should raise an exception here, if the `method_to_compute `is not amongst the predefined ones. I don't see the point of return an empty dict, since it will fail a bit after when calling `score_dict[score_method]` ?

Comment (natct10) [reply contributivity_measures.py:62]: Good point!

Comment (celinejacques) [reply contributivity_measures.py:62]: @RomainGoussault @natct10 But in fact this exception is raised during the initialization of scenario. 
- If a method is not in the default list or if there are no method in the list given in the config file, the script ends. 
- If there is no method field in config file, it consedered that it is the default situation, so it runs all methods.

Comment (Thomas-Galtier) [reply contributivity_measures.py:62]: @celinejacques  Your second point is not right, you changed the code so it would rather exit the script ;)

Comment (natct10) [reply contributivity_measures.py:62]: Okay, great!

File: simulation_run.py [hunk 5]
```diff
@@ -101,73 +101,23 @@ def run_scenario(current_scenario):
...
+        
+        for score_method in score_dict:
+            score = score_dict[score_method]
+            contrib = contributivity.Contributivity(
+                score_method, score[0], score[1], np.round(end - start)
+                )
+    
+            current_scenario.append_contributivity(contrib)
+            print("\n## Evaluating contributivity with " + method + ":")
+            print(contrib)
```
bowni, line 119: @celinejacques if I follow correctly, `compute_contributivity` is called method per method, and so `score_dict` receives only the scores for one and only one method. So if this is correct, then I think we don't need to iterate over `score_dict` with the `for score_method in score_dict:` loop. Or did I misunderstood something?

celinejacques: @bowni We need it for the method `independant score` as it gives two score: `raw` and `additive`.

scenario.py
```diff
@@ -71,10 +72,29 @@ def __init__(self, params, experiment_path):
...
+            "Shapley values",
+            "Independant scores",
+            "TMCS values"
+        ]
+        
+        self.methods = []
+        if not 'methods' in params:
+            self.methods = methods_default
+        else:
+            if not params['methods']:
```
RomainGoussault, line 90: Why do we need this check ? Is it the same check as in line 87 ?

scenario.py
```diff
@@ -71,10 +72,29 @@ def __init__(self, params, experiment_path):
...
             self.epoch_count = 40
+            
+        methods_default = [
+            "Shapley values",
+            "Independant scores",
+            "TMCS values"
+        ]
+        
+        self.methods = []
+        if not 'methods' in params:
```
RomainGoussault, line 87: I would start with the positiv case for simplicity : ` if 'methods' in params:`

scenario.py
```diff
@@ -71,10 +72,29 @@ def __init__(self, params, experiment_path):
...
+            self.methods = methods_default
+        else:
+            if not params['methods']:
+                sys.exit('No contributivity method given in config file')
+            else:
+                for el in params['methods']:
+                    if el in methods_default:
+                        self.methods.append(el)
+                    else:
+                        sys.exit('method ' + el + ' is not in methods list.')
```
RomainGoussault, line 97: I would raise an exception here instead of exit since this is not a "normal" exit:
`raise Exception("No contributivity method given in config file")`

celinejacques: Yes, that's probably better! I changed it.

scenario.py
celinejacques: @RomainGoussault 
line 87: we check if the field `methods` is given in the config file.
line 90: if the `methods` field is given in config file, we check if it is empty or not.

simulation_run.py
bowni: Ok got it. So it gives the possibility in a contributivity measurement method to produce multiple scores, that is rather a good thing then.

bowni:
I fixed the merge conflicts.

scenario.py
```diff
@@ -77,24 +78,38 @@ def __init__(self, params, experiment_path):
...
+            self.methods = methods_default
+        else:
+            if params['methods']:
+                for el in params['methods']:
+                    if el in methods_default:
+                        self.methods.append(el)
+                    else:
+                        raise Exception('method ' + el + ' is not in methods list.')
+            else:
+                raise Exception("No contributivity method given in config file")
```
RomainGoussault, line 111: What do you think of something like this:
```
ALL_METHODS_LIST =  = [...]
DEFAULT_METHODS_LIST = [...]

if 'methods' in params:

	for method in params['methods']:
			if method in ALL_METHODS_LIST:
				self.methods.append(method)
			else:
				raise Exception('method ' + method + ' is not in methods list: ' + ALL_METHODS_LIST)
				
else:
	self.methods = DEFAULT_METHODS_LIST
```

I choose to put 2 methods list (DEFAULT and ALL) because I guess we might not want to run all methods by default.

Thomas-Galtier: I did the same on the IS-method PR

bowni:
@celinejacques can you check with the Travis' build fail? Follow the 'Details' link below, next to the line indicating it fails. There seem to be something that doesn't execute properly.

bowni:
> @celinejacques can you check with the Travis' build fail? Follow the 'Details' link below, next to the line indicating it fails. There seem to be something that doesn't execute properly.

And they pass now, well done!

scenario.py
```diff
@@ -77,24 +78,47 @@ def __init__(self, params, experiment_path):
...
+            self.methods = methods_default
+        else:
+            if params['methods']:
+                for el in params['methods']:
+                    if el in known_methods_list:
+                        self.methods.append(el)
+                    else:
+                        raise Exception('method ' + el + ' is not in methods list.')
+            else:
+                raise Exception("No contributivity method given in config file")
```
RomainGoussault, line 118: What do you think of something like this:
```
ALL_METHODS_LIST =  = [...]
DEFAULT_METHODS_LIST = [...]

if 'methods' in params:

	for method in params['methods']:
			if method in ALL_METHODS_LIST:
				self.methods.append(method)
			else:
				raise Exception('method ' + method + ' is not in methods list: ' + ALL_METHODS_LIST)
				
else:
	self.methods = DEFAULT_METHODS_LIST
```

I choose to put 2 methods list (DEFAULT and ALL) because I guess we might not want to run all methods by default.

bowni: To me that's what is currently implemented (except the `if ... else ...` is reversed but it doesn't change anything): `DEFAULT_METHODS_LIST` is `methods_default` and equivalent to your `ALL_METHODS_LIST` is `known_methods_list`. Or what do I miss?

Thomas-Galtier: I changed the code so it better fits Romain's suggestion

celinejacques: @bowni Yes it is a same but in a more elegant way ;) 
@RomainGoussault I update the code! What do you want in `DEFAULT_METHODS_LIST` ? For the moment, do I write the 3 methods we currently have?

celinejacques: @RomainGoussault But in fact, in your code, we do not check if a list is given in config file. If we have the field method but nothing is given, it stops running.

PR #90, opened 2020-04-02
Add new local_perf option for aggregation after each local training iteration

bowni:
Add new local_perf option for aggregation after each local training iteration


bowni:
I rebased with `iss4-mini-batches` and I probably shouldn't have... So I created a new branch, cherry-picked the 2 significant commits from it, and will create a new PR. Closing this one.

PR #93, opened 2020-04-02
Add option for aggregation weighted per local performance

### LINKED ISSUE #88: Add the possibility to weight the aggregation with local performances of the last mini-batch
This is to have a second weighting option as an alternative to weighting per data volume. This alternative would be to weight the averaging aggregation with the performance of each node's individual model on the global validation set.

bowni:
Closes #88.

### DIFF
--- a/fl_training.py
+++ b/fl_training.py
@@ -167,7 +167,7 @@ def prepare_aggregation_weights(aggregation_weighting, nodes_count, node_list, i
     elif aggregation_weighting == "data_volume":
         node_sizes = [len(node.x_train) for node in node_list]
         aggregation_weights = node_sizes / np.sum(node_sizes)
-    elif aggregation_weighting == "local_perf":
+    elif aggregation_weighting == "local_score":
         aggregation_weights = input_weights / np.sum(input_weights)
     else:
         raise NameError(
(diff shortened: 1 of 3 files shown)


fl_training.py
```diff
@@ -315,7 +318,10 @@ def compute_test_score(
     # After last epoch or if early stopping was triggered, evaluate model on the global testset
     print("\n### Evaluating model on test data:")
     model_evaluation = aggregated_model.evaluate(
-        x_test, y_test, batch_size=constants.BATCH_SIZE, verbose=0
+        x_test,
+        y_test,
+        batch_size=constants.BATCH_SIZE,
+        verbose=0,
```
natct10, line 324: Thx!

fl_training.py
```diff
@@ -257,6 +257,7 @@ def compute_test_score(
 
             # Starting model for each node is the aggregated model from the previous mini-batch iteration
             agg_model_for_iteration = [None] * nodes_count
+            aggregation_weights = prepare_aggregation_weights(aggregation_weighting, nodes_count, node_list, local_perf_list)
```
RomainGoussault, line 260: We should save the `local_perf_list `in the same way we save `global_val_acc `and `global_val_loss`. This way we would be able to have an insight of how the perf and therefore the weights change/evolve during training.

fl_training.py
```diff
@@ -158,7 +158,7 @@ def split_in_minibatches(minibatch_count, x_train, y_train):
 
     return minibatched_x_train, minibatched_y_train
 
-def prepare_aggregation_weights(aggregation_weighting, nodes_count, node_list):
+def prepare_aggregation_weights(aggregation_weighting, nodes_count, node_list, input_weights):
```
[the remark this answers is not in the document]
bowni: I gave a generic name to this argument because we might developed additional weighting approaches for which we need to pass to this method a list of input weights. So I think it is cleaner to do that way (instead of creating rigidity for a specific case like the weighting per local perf). Is this OK with you @RomainGoussault ?

fl_training.py
```diff
@@ -167,6 +167,8 @@ def prepare_aggregation_weights(aggregation_weighting, nodes_count, node_list):
     elif aggregation_weighting == "data_volume":
         node_sizes = [len(node.x_train) for node in node_list]
         aggregation_weights = node_sizes / np.sum(node_sizes)
+    elif aggregation_weighting == "local_perf":
```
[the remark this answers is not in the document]
bowni: Ok, done

fl_training.py
bowni: Ok, done

bowni:
@RomainGoussault I implemented the changes you requested except 1, see comments. For the saving of scores I created a new score matrix: `score_matrix_extended` because I didn't want to mess will all the usages of the existing `score_matrix`. I let you play with it and adapt downstream code, if relevant.

RomainGoussault:
@bowni code looks good to me 👍.
I pushed one commit and i am currently running simulations to test the changes!

PR #100, opened 2020-04-08
Update visualization notebook

RomainGoussault:
It's probably best to look at the file here https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/c757c1b26afd17b026941d7b0238320ba406e451/analyse_results.ipynb
The diff for notebooks does not work well...

This work is just a base to build up more advanced figures later. 

Also the results file used had only one random state done, that is why I used swarmplot. Boxplot would be nicer when we have more iterations.


natct10:
Great work @RomainGoussault ! I would just tweak the `scenario_name` which also contains `scenario_id` :
```sh
scenario_id: 2
scenario name: Id 2 - Stratified[0.4, 0.3, 0.3] mb 20.0
```
Also, is there a chance to fix the overlapping method names in the graphs (on the X axis)?

Thomas-Galtier:
These results show  there is a bug with SMC methods obviously

RomainGoussault:
@natct10 The scenario name is not fixed, it should change for every experiment depending on what we want to do.
Concerning the overlapping method names, we could use a smaller font, or rotate them or find shorter names @Thomas-Galtier :) ?

natct10:
> shorter names

:+1:

RomainGoussault:
I have created an issue for shorter name https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/104 @Thomas-Galtier

PR #107, opened 2020-04-16
early stopping for single node

Thomas-Galtier:


fl_training.py
```diff
@@ -82,14 +83,27 @@ def compute_test_score_for_single_node(
...
+        )   
+    else:
+        history = model.fit(
+            node.x_train,
+            node.y_train,
+            batch_size=constants.BATCH_SIZE,
+            epochs=epoch_count,
+            verbose=0,
+            validation_data=(node.x_val, node.y_val),
+        )
```
bowni, line 105: Just a question because I am wondering: wouldn't it be cleaner to put the `if ... else ...` statement in the content of a variable `callback_list = ...` instead of the whole `model.fit()`?

RomainGoussault: I agree !

Issue #92, opened 2020-04-02
Parameterize the BATCH_SIZE constant


bowni:
See [this issue on StackExchange](https://stats.stackexchange.com/questions/153531/what-is-batch-size-in-neural-network).

I made a few tests in our setup (on my local machine...) with small and large `batch_size`:

- computation time goes up when too small
- a `batch_size` that is too large makes the model convergence really slow
- it looks like somewhere near ~10 batches is a good compromise

PR #116, opened 2020-04-23
Enable launching only the distributed learning and parameterize BATCH_SIZE

### LINKED ISSUE #87: Add param to only run federated scenario without contributivity methods
Building on https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/78 (Add params to choose which contributivity methods to run), sometimes we only to need to run the federated training without computing contributivity measures. For instance when we want to test the federated implementation in particular case.

Note: for now, a quick hack would be to only compute, one cheap contributivity measure (like the independant contributivity method)

bowni:
Closes #87 #92 #114


bowni:
I made a schema for explaining the different iteration levels: `epoch`, `mini-batch`, `fit batch`. I thought we could add it to the `README.md`. What do you think? (it is a shared drawing you can [edit it here](https://docs.google.com/drawings/d/1bTJRhBKN9RpcEVfimWURHbbGVl8m6ICSmk1ryZHTVbY/edit?usp=sharing))

epoch_minibatch_fitbatch

bowni:
@RomainGoussault I merged master into the advanced split branch, and then the advanced split branch in this one. I started converting prints to proper logging. There is one issue I can't find a solution to:
I had tuned `fl_training` prints during a mini-batch to print on the same line (using `end="", flush=True` as arguments passed to `print()`). It seems that the way the logging approach picks the prints and renders them to the console doesn't pick that up. And I can't find how to do that by doing a proper `logger.info()` to those elements. Any idea?

RomainGoussault:
> @RomainGoussault I merged master into the advanced split branch, and then the advanced split branch in this one. I started converting prints to proper logging. There is one issue I can't find a solution to:
> I had tuned `fl_training` prints during a mini-batch to print on the same line (using `end="", flush=True` as arguments passed to `print()`). It seems that the way the logging approach picks the prints and renders them to the console doesn't pick that up. And I can't find how to do that by doing a proper `logger.info()` to those elements. Any idea?


Good question, I don't have a good answer.. we could ask loguru, they seem to be pretty reactive: https://github.com/Delgan/loguru/issues

README.md
```diff
@@ -190,24 +200,24 @@ Example: `epoch_count: 30`
 The distributed learning approach implemented relies on a sequence of (i) training in parallel on each partner's dataset, and then (ii) aggregating the resulting models. This constitutes an individual iteration. These iterations are repeated for all _mini-batches_ into which the partner's datasets are split at the beginning of each epoch. This gives a total of `epoch_count* minibatch_count` iterations.  
 Example: `minibatch_count: 20`
 
+`fit_batches_count`: `int` (default: `8`)  
+The ML training implemented relies on Keras' `.fit()` function, which takes as argument a `batch_size`, used as the number of samples per gradient update. Depending on the number of samples in the train dataset, this defines how many gradient updates are done per `.fit()` iteration. The `fit_batches_count` parameter enables to specify this number of gradient updates per `.fit()` iteration (both in multi-partner setting where there is 1 `.fit()` iteration per mini-batch, and in single-partner setting where there is 1 `.fit()` iteration per epoch).
```
RomainGoussault, line 204: > both in multi-partner setting where there is 1 `.fit()` iteration per mini-batch, and in single-partner setting where there is 1 `.fit()` iteration per epoch

Is it really what you meant ?
And in single partner setting, there is one fit() call for all the epochs

fl_training.py
```diff
@@ -356,7 +342,7 @@ def compute_test_score(
     # After last epoch or if early stopping was triggered, evaluate model on the global testset
     print("\n### Evaluating model on test data:")
     model_evaluation = aggregated_model.evaluate(
-        x_test, y_test, batch_size=constants.BATCH_SIZE, verbose=0,
+        x_test, y_test, verbose=0,
```
RomainGoussault, line 345: So we use the default batch size of 32 ?

main.py
```diff
@@ -71,7 +71,7 @@ def main():
                 logger.info(scenario_params)
 
                 current_scenario = scenario.Scenario(scenario_params, experiment_path)
-                print(current_scenario.to_dataframe())
+                # print(current_scenario.to_dataframe())  # to be deleted?
```
RomainGoussault, line 74: yes

scenario.py
```diff
@@ -553,6 +561,6 @@ def to_dataframe(self):
 
                 df = df.append(dict_results, ignore_index=True)
 
-        df.info()
+        # df.info()  # to be deleted?
```
RomainGoussault, line 564: yes

scenario.py
```diff
@@ -472,6 +469,16 @@ def plot_data_distribution(self):
         # plt.show()  # DEBUG
         plt.savefig(self.save_folder / "data_distribution.png")
 
+    def compute_batch_sizes(self):
+
+        # For each partner we compute the batch size in multi-partner and single-partner setups
+        for p in self.partners_list:
+            p.batch_size_multi = max(1, int(len(p.x_train) / (self.minibatch_count * self.fit_batches_count)))
+            p.batch_size_single = max(1, int(len(p.x_train) / self.fit_batches_count))
+            logger.debug(f"   compute_batch_sizes(), partner #{p.id}: "
```
RomainGoussault, line 478: I think we can log at level INFO here ?

scenario.py
```diff
@@ -103,6 +104,12 @@ def __init__(self, params, experiment_path):
         else:
             self.minibatch_count = 20
 
+        if "fit_batches_count" in params:
+            self.fit_batches_count = params["fit_batches_count"]
+            assert self.fit_batches_count > 0
```
RomainGoussault, line 109: so this the number of gradient update per mini batch ?
Could we call it `gradient_updates_per_minbach_count `? It's a bit long, but it's hard to guess what is `fit_batches_count`

partner.py
```diff
@@ -12,10 +12,14 @@
 
 
 class Partner:
+
     def __init__(self, partner_id):
 
         self.id = partner_id
 
+        self.batch_size_single = constants.DEFAULT_BATCH_SIZE
+        self.batch_size_multi = constants.DEFAULT_BATCH_SIZE
```
RomainGoussault, line 21: batch_size_single_partner / batch_size_multi_parnter

RomainGoussault: Do we have to assign a value to them here since `compute_batch_sizes`() will overwrite these values ?

scenario.py
bowni: Ok, I'll add it

README.md
bowni: I'll clarify

scenario.py
bowni: I propose to keep it that way, with its definition + explicative schema in the `README.md`, since it resonates with `batch_size` and is specific to usage of `fit()`. If in a few weeks it still looks odd we'll change it.

README.md
```diff
@@ -187,27 +197,27 @@ Number of epochs passed as argument in the `.fit()` function. Superseded when `i
 Example: `epoch_count: 30`
 
 `minibatch_count`: `int` (default: `20`)  
-The distributed learning approach implemented relies on a sequence of (i) training in parallel on each partner's dataset, and then (ii) aggregating the resulting models. This constitutes an individual iteration. These iterations are repeated for all _mini-batches_ into which the partner's datasets are split at the beginning of each epoch. This gives a total of `epoch_count* minibatch_count` iterations.  
+The distributed learning approach implemented relies on a sequence of (i) training in parallel on each partner's dataset, and then (ii) aggregating the resulting models. This constitutes an individual iteration. These iterations are repeated for all *mini-batches* into which the partner's datasets are split at the beginning of each epoch. This gives a total of `epoch_count * minibatch_count` iterations.  
 Example: `minibatch_count: 20`
 
+`fit_batches_count`: `int` (default: `8`)
```
RomainGoussault, line 203: I still find this variable name impossible to understand (without the explanation)

config.yml
```diff
@@ -30,6 +34,10 @@ scenario_params_list:
    epoch_count: 
      - 38
    methods:
-     - ["Shapley values", "Independant scores", "TMCS"]
+     - "Shapley values"
```
RomainGoussault, line 37: Need to put them inside a list (like for the quick debug config) otherwise it crashes.

Actually I think it will make more sense, if we could write the methods like this (and not have to put them into a list) but this can be the matter of another PR...

bowni: OK

README.md
bowni: OK, you win

constants.py
```diff
@@ -12,7 +12,7 @@
 
 # ML constants
 DEFAULT_BATCH_SIZE = 256
-DEFAULT_FIT_BATCHES_COUNT = 8
+DEFAULT_gradient_updates_per_pass_count = 8
```
RomainGoussault, line 15: should be all UPPER CASE :) !

PR #121, opened 2020-04-24
Logging

### LINKED ISSUE #32: Save print into log files

RomainGoussault:
Fixes #32 
Create two logs files : 
- one for logs level INFO and above
- one for logs level DEBUG and above

Log files are created at the root folder. When the experiment folder is created, logs files are moved to this folder.

stdout and exceptions are logged. So any print() is logged but we would have to change them to logger.info() or logger.debug() / warning() ... for better clarity --> https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/123
Tensorflow messages are not logged. I created an issue for that --> https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/122

Logging is done with Loguru : https://github.com/Delgan/loguru

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -17,5 +17,9 @@ PATIENCE = 4  # patience for early stopping
 # GPU
 GPU_MEMORY_LIMIT_MB = 4096
 
+# Logging
+INFO_LOGGING_FILE_NAME = "info.log"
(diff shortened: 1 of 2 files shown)


RomainGoussault:
> Great!
> Question: do logger.debug() prints actually print on the console?

Yes !

PR #124, opened 2020-04-24
Iss96 output simulation run

### LINKED ISSUE #96: Improve output of simulation_run
Now, the output of the simulation_run.py script is:

- 1 csv file
- many txt files
- many .npy files which contains the loss/acc during training 

A better situation would be:

- 1 csv file
- 1 numpy or binary file which contains the loss/acc during training (we may be able to put it also in the csv file)
- 1 html/pdf page generated by a standard jupyter notebook to look at basic results.

celinejacques:
closes #96

### DIFF
--- a/fl_training.py
+++ b/fl_training.py
@@ -6,6 +6,8 @@ Functions for model training and evaluation (single-partner and multi-partner ca
 
 from __future__ import print_function
 
+import os
+import pickle
 import keras
 from keras.backend.tensorflow_backend import clear_session
 from keras.callbacks import EarlyStopping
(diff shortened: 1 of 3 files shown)


fl_training.py
```diff
@@ -375,24 +368,29 @@ def compute_test_score(
     if is_save_fig:
 
         # Save data
-        np.save(save_folder / "score_matrix", score_matrix)
-        np.save(save_folder / "global_val_acc", global_val_acc)
-        np.save(save_folder / "global_val_loss", global_val_loss)
-        np.save(save_folder / "score_matrix_extended", score_matrix_extended)
-
+        metrix_results = {}
```
bowni, line 371: I have 2 questions on that part:

1. I like the sound of "metrix", it is quite cool :-) But still, I think we should write "metrics"?

1. What exactly does the following bit and how can `metrix_results.p` be used for exploring results?
```python
with open(save_folder / "metrix_results.p", 'wb') as f:
            pickle.dump(metrix_results, f)
```

fl_training.py
```diff
@@ -375,24 +368,29 @@ def compute_test_score(
...
-        np.save(save_folder / "global_val_acc", global_val_acc)
-        np.save(save_folder / "global_val_loss", global_val_loss)
-        np.save(save_folder / "score_matrix_extended", score_matrix_extended)
-
+        metrics_results = {}
+        metrics_results["score_matrix"] = score_matrix
+        metrics_results["global_val_acc"] = global_val_acc
+        metrics_results["global_val_loss"] = global_val_loss
+        metrics_results["score_matrix_extended"] = score_matrix_extended
+        with open(save_folder / "metrics_results.p", 'wb') as f:
```
RomainGoussault, line 376: Since we have both metrics and loss, I would name the file something like history_data (it's the name used by Keras)

PR #125, opened 2020-04-24
Only one method per scenario

celinejacques:
Now, only one method is processed by scenario.

### DIFF
--- a/config_quick_debug.yml
+++ b/config_quick_debug.yml
@@ -13,7 +13,7 @@ scenario_params_list:
     - [not_corrupted, not_corrupted, corrupted]
    is_quick_demo: 
     - True
-   methods:
+   method:
     - "Independant scores"
     - "Independant scores"
 
--- a/config_quick_debug.yml
+++ b/config_quick_debug.yml
@@ -14,5 +14,6 @@ scenario_params_list:
    is_quick_demo: 
     - True
    methods:
-    - ["Independant scores"]
+    - "Independant scores"
+    - "Independant scores"
 
--- a/scenario.py
+++ b/scenario.py
@@ -40,7 +40,7 @@ class Scenario:
         self.contributivity_list = []
 
         # -------------------------------------
-        # Definition of collaborative scenarios
+        # Definition of collaborative scenarios
         # -------------------------------------
 
         # partners mock different partners in a collaborative data science project
@@ -66,7 +66,7 @@ class Scenario:
             self.corrupted_datasets = ["not_corrupted"] * self.partners_count
 
         # ---------------------------------------------------
-        #  Configuration of the distributed learning approach
+        #  Configuration of the distributed learning approach
         # ---------------------------------------------------
 
         # When training on a single partner, the test set can be either the local partner test set or the global test set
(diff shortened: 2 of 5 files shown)


config_quick_debug.yml
```diff
@@ -13,6 +13,7 @@ scenario_params_list:
     - [not_corrupted, not_corrupted, corrupted]
    is_quick_demo: 
     - True
-   methods:
-    - ["Independant scores"]
+   method:
+    - "Independant scores"
+    - "Independant scores"
```
RomainGoussault, line 18: One should be enough ?

scenario.py
```diff
@@ -122,19 +122,15 @@ def __init__(self, params, experiment_path):
...
-
-        self.methods = []
-        if "methods" in params and params["methods"]:
-
-            for method in params["methods"]:
-                if method in ALL_METHODS_LIST:
-                    self.methods.append(method)
-                else:
-                    raise Exception("Method [" + method + "] is not in methods list.")
+        DEFAULT_METHOD= "Independant scores"
```
RomainGoussault, line 125: DEFAULT_METHOD =

bowni:
> I think there are two main issues we need to tackle:
> 
>     * If we have 3 methods, then we compute three times the federated training `fl_training.compute_test_score_with_scenario()` for nothing.

That is a good point.
Actually **we don't _have_ to** compute the multi-partner collaborative training when we want to compare contributivity measurements. I think we just took the habit of computing it every time we launch an experiment, **because we became interested** in the score of the multi-partner collaborative training.
This thing forces us to take one step back and re-think the Experiment and Scenario definitions / roles. We might need a third level. For example Experiment > Scenario > Comparisons:

- Experiment: a set of different scenarios
- Scenario: a given multi-partner scenario (dataset, data amounts, split option, etc.)
  - Learning Comparison: different multi-partner collaborative learning methods applied to the same scenario
  - Contributivity Measurement Comparison: different contributivity measurement methods (for a given learning method passed as parameter)

Probably it would make more sense to save results of Learning Comparisons and Contributivity Measurement Comparisons in different result files.

Does this seem to make sense?

> 
>     * If we have 3 methods, then we will have 3 scenario ids in the result file. The result analysis will be way harder if we need to reconstruct a posteriori the "original" scenario id.

In the approach described above we would have:

- Each 'Learning Comparison' would have a dedicated id under the umbrella of the same Experiment id + Scenario id
- Each 'Contributivity Measurement Comparison' would have a dedicated id under the umbrella of the same Experiment id + Scenario id

bowni:
So now that we have implemented the extra multi-partner learning methods, we have a clearer view of what really makes sense. And it turns out the current implementation is pretty cool: in a given scenario there is 1 multi-partner learning approach and multiple contributivity methods. This is what is most natural (I now think).
I would say, as a consequence, that we should probably drop this ongoing PR.

bowni:
> So now that we have implemented the extra multi-partner learning methods, we have a clearer view of what really makes sense. And it turns out the current implementation is pretty cool: in a given scenario there is 1 multi-partner learning approach and multiple contributivity methods. This is what is most natural (I now think).
> I would say, as a consequence, that we should probably drop this ongoing PR.

Is it ok @celinejacques @RomainGoussault that would stop and delete this PR?

RomainGoussault:
> > So now that we have implemented the extra multi-partner learning methods, we have a clearer view of what really makes sense. And it turns out the current implementation is pretty cool: in a given scenario there is 1 multi-partner learning approach and multiple contributivity methods. This is what is most natural (I now think).
> > I would say, as a consequence, that we should probably drop this ongoing PR.
> 
> Is it ok @celinejacques @RomainGoussault that would stop and delete this PR?

I agree

Issue #127, opened 2020-04-30
Something breaks when launching a scenario with only 1 partner


bowni:
```
  File "C:\Users\Eric\Dev\Substra\distributed-learning-contributivity\scenario.py", line 370, in split_data
    splitting_indices[0] = self.amounts_per_partner[0]
    │                      │    └ [1]
    │                      └ <scenario.Scenario object at 0x000002921CAD3860>
    └ array([], dtype=float64)

IndexError: index 0 is out of bounds for axis 0 with size 0
```

PR #129, opened 2020-05-05
Fix a bunch of bugs and simple issues

### LINKED ISSUE #120: Too many figures opened
I add this warning when I tested a case with 10 scenarios:

    /home/eric/dev/distributed-learning-contributivity/fl_training.py:389: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (matplotlib.pyplot.figure) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam figure.max_open_warning).

@RomainGoussault you might be the one the most familiar with this, should we do something or is it OK?

bowni:
Fixes #120, #127, #123, #130

### DIFF
--- a/contributivity.py
+++ b/contributivity.py
@@ -204,7 +204,7 @@ class Contributivity:
         end = timer()
         self.computation_time_sec = end - start
 
-    # %% compute Shapley values with the truncated Monte-carlo metho
+    # %% compute Shapley values with the truncated Monte-carlo method
     def truncated_MC(self, the_scenario, sv_accuracy=0.01, alpha=0.9, truncation=0.05):
         """Return the vector of approximated Shapley value corresponding to a list of partner and a characteristic function using the truncated monte-carlo method."""
         start = timer()
--- a/contributivity.py
+++ b/contributivity.py
@@ -131,7 +131,7 @@ class Contributivity:
                     subset_with_i = np.sort(np.append(subset, i))
                     if (
                         tuple(subset_with_i) in self.charac_fct_values
-                    ):  # we store the new knwon increments
+                    ):  # we store the new known increments
                         self.increments_values[i][tuple(subset)] = (
                             self.charac_fct_values[tuple(subset_with_i)]
                             - self.charac_fct_values[tuple(subset)]
@@ -890,7 +890,6 @@ class Contributivity:
                 np.any(continuer) or (1 - alpha) < v_max / (sv_accuracy ** 2)
             ):  # Check if the length of the confidence interval  is below the value of sv_accuracy
                 t += 1
-                # print(t)  # to be deleted?
                 for k in range(N):
                     # select the strata to add an increment
                     if np.any(continuer[k]):
--- a/contributivity.py
+++ b/contributivity.py
@@ -12,6 +12,7 @@ from scipy.stats import norm
 from itertools import combinations
 from math import factorial
 from sklearn.linear_model import LinearRegression
+from loguru import logger
 
 import fl_training
 import shapley_value.shapley as sv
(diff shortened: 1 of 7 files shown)


scenario.py
```diff
@@ -468,12 +471,19 @@ def plot_data_distribution(self):
...
         # plt.show()  # DEBUG
         plt.savefig(self.save_folder / "data_distribution.png")
+        plt.close()
 
     def compute_batch_sizes(self):
 
         # For each partner we compute the batch size in multi-partner and single-partner setups
+        if self.partners_count == 1:
+            p = self.partners_list[0]
+            p.batch_size = max(1, int(len(p.x_train) / self.gradient_updates_per_pass_count))
```
RomainGoussault, line 481: Why don't we use the same batch size calculation when there is only one partner ?

bowni: If there is only one partner in the scenario there is not going to be any multi-partner learning at all, only single partner learning cases. As a consequence there is no minibatch iterations, so it would not make sense to compute the `batch_size` based on the number of gradient updates per pass for the multi-partner learning cases. In this (rare) case, we compute the `batch_size` for enforcing the `gradient_updates_per_pass_count` in single partner learning case.

contributivity.py
```diff
@@ -890,17 +885,17 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
 
-            # sampnling
+            # Sampling
             while (
-                np.any(continuer) or (1 - alpha)  <  v_max / (sv_accuracy ** 2)
+                np.any(continuer) or (1 - alpha) < v_max / (sv_accuracy ** 2)
             ):  # Check if the length of the confidence interval  is below the value of sv_accuracy
                 t += 1
-                print(t)
+                # print(t)  # to be deleted?
```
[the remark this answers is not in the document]
bowni: Done

utils.py
```diff
@@ -111,8 +111,8 @@ def get_scenario_params_list(config):
                     raise Exception("Length of corrupted_datasets does not match number of partners.")
 
             scenario_params_list.append(scenario)
-            
-    print('\nNumber of scenario(s) configured:', len(scenario_params_list))
+
+    logger.info(f"Number of scenario(s) configured: {len(scenario_params_list)}")
```
natct10, line 115: Looks good, I will review tomorrow but it seems to be a nice cleanup, thank you @bowni

contributivity.py
```diff
@@ -142,31 +143,27 @@ def not_twice_characteristic(self, subset, the_scenario):
 
     def compute_SV(self, the_scenario):
         start = timer()
-        print("\n# Launching computation of Shapley Value of all partners")
+        logger.info("# Launching computation of Shapley Value of all partners")
```
natct10, line 146: :+1:

main.py
```diff
@@ -67,11 +65,11 @@ def main():
 
             for scenario_id, scenario_params in enumerate(scenario_params_list):
 
+                logger.info(f"Scenario {scenario_id + 1}/{len(scenario_params_list)}")
```
natct10, line 68: Naive question: is there any possible case where the len of `scenario_params_list` could be equal to 0?

main.py
```diff
@@ -140,12 +138,11 @@ def run_scenario(current_scenario):
     current_scenario.federated_computation_time_sec = end - start
 
     for method in current_scenario.methods:
-        print(method)
+        logger.info(f"{method}")
```
natct10, line 141: Maybe this one is not required as it is logged just after?

scenario.py
RomainGoussault: My bad, I thought it was when doing contributivity computations on a subset of partners.
This is indeed a rare case and this not distributed learning but "standard" machine learning.

main.py
bowni: Agreed. Actually I changed it in PR #131

main.py
bowni: Hmm, no I don't think so. It would mean launching the program with a config file containing no scenario: doesn't make sense.

PR #131, opened 2020-05-05
Additional multi partner learning approaches

### LINKED ISSUE #91: Add param to choose between parallel or sequential federated learning
From the Substra whitepaper: 
image
We now only do parallel FL, we could a param to do sequential FL.

bowni:
Fixes #91


scenario.py
```diff
@@ -63,66 +63,82 @@ def __init__(self, params, experiment_path):
...
+            "seq",
+            "seqavg",
+        ]
+
+        if "multi_partner_learning_approach" in params:
+            approach = params["multi_partner_learning_approach"]
+            if approach in multi_partner_learning_approaches_list:
+                self.multi_partner_learning_approach = approach
+            else:
+                raise Exception(f"Multi-partner learning approach '{approach}' is not in methods list.")
```
RomainGoussault, line 90: is not in **multi_partner_learning_approaches_list** list.
Actually somethinkg like "Multi-partner learning approach '{approach}' is not a valid approach" might be clearer

fl_training.py
```diff
@@ -205,13 +205,266 @@ def build_aggregated_model(new_weights):
...
+        batch_size=batch_size,
+        epochs=1,
+        verbose=0,
+        validation_data=val_data,
+    )
+
+    return history
+
+
+def print_collaborative_round_partner_result(
```
RomainGoussault, line 268: log_collaborative... ?

fl_training.py
```diff
@@ -205,13 +205,266 @@ def build_aggregated_model(new_weights):
...
+
+    # Initialize variables
+    partners_count = len(partners_list)
+    epoch_index, epoch_count, minibatch_index, minibatch_count = collaborative_round_indexes
+    minibatched_x_train, minibatched_y_train = train_data
+
+    # Iterate over partners for training the model sequentially
+    shuffled_indexes = np.random.permutation(len(partners_list))
+    logger.debug(f"(seq) Shuffled order for this sequential collaborative round: {shuffled_indexes}")
+    for for_loop_idx, partner_index in enumerate(shuffled_indexes):
```
RomainGoussault, line 374: What is `for_loop_idx `??

fl_training.py
```diff
@@ -205,13 +205,266 @@ def build_aggregated_model(new_weights):
...
+    val_acc_str = f"{round(validation_score, 2)}"
+
+    logger.info(f"{epoch_nb_str} > {mb_nb_str} > {partner_id_str} > val_acc: {val_acc_str}")
+
+
+def update_iterative_results(partner_index, collaborative_round_indexes, fit_history, iterative_results):
+    """Update the results arrays with results from the collaboration round"""
+
+    epoch_index, epoch_count, minibatch_index, minibatch_count = collaborative_round_indexes
+    local_score_list, model_list, score_matrix, score_matrix_extended = iterative_results
```
RomainGoussault, line 286: If it makes things simpler, all of these variable could be members of the fl_scenario class. So we won't have to pass them around functions like we have to do now.
What do you think ?

scenario.py
bowni: OK

fl_training.py
bowni: OK

fl_training.py
bowni: The index of iterations over the for loop

fl_training.py
bowni: Oh yes, much better. But we don't have a class. Should I create one?

scenario.py
bowni: Done

fl_training.py
bowni: Done

fl_training.py
RomainGoussault: Yes ! We can transform fl_scenario to be a class.

PR #138, opened 2020-05-15
Small improvements on the go

bowni:


RomainGoussault:
Nice schematic ! I would just modify the yellow color which is a bit hard to see.

Issue #135, opened 2020-05-12
Save the weights instead of the models


bowni:
While working on #131 I thought that we could save only the weights of the models instead of the models themselves in `model_list`. The reason we save them is to later extract their weights for aggregation purposes. So it doesn't seem optimal. What do you think @RomainGoussault ?

RomainGoussault:
Yes I agree, if we are only aggregating the weights there is no point saving the model itself. Good point !

PR #141, opened 2020-05-19
Refactor multi partner learning

### LINKED ISSUE #134: Refactor fl_training
fl_training module should probably be a class. This would make the code more readable.

bowni:
Fixes #134, #135, #136, #145.


multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+        aggregated_model.compile(
+            loss=keras.losses.categorical_crossentropy,
+            optimizer="adam",
+            metrics=["accuracy"],
+        )
+
+        return aggregated_model
+
+    @staticmethod
+    def build_from_previous_model(previous_model):
```
RomainGoussault, line 437: Where is this used ?

multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+
+        # At the end of each mini-batch, for each partner, populate the extended score matrix
+        self.score_matrix_extended[self.epoch_index, self.minibatch_index, partner_index] = validation_score
+
+        # At the end of each epoch (at end of last mini-batch), for each partner, populate the score matrix
+        if self.minibatch_index == (self.minibatch_count - 1):
+            self.score_matrix[self.epoch_index, partner_index] = validation_score
+
+
+def init_multipartnerlearning_from_scenario(scenario, is_save_fig=False):
```
RomainGoussault, line 514: what do you think @natct10 ?

main.py
```diff
@@ -129,14 +129,13 @@ def run_scenario(current_scenario):
...
+    current_scenario = utils.preprocess_scenarios_data(current_scenario)
 
     # Train and eval on all partners according to scenario
     is_save_fig = True
     start = timer()
-    current_scenario.federated_test_score = fl_training.compute_test_score_with_scenario(
-        current_scenario, is_save_fig
-    )
+    mpl = multi_partner_learning.init_multipartnerlearning_from_scenario(current_scenario)
+    current_scenario.federated_test_result = mpl.compute_test_score()
```
RomainGoussault, line 138: mpl could be an attribute of the scenario instance. This way it would be easier to get back results from the mpl instance to the scenario instance.

scenario.py
```diff
@@ -532,7 +532,8 @@ def to_dataframe(self):
         dict_results["minibatch_count"] = self.minibatch_count
         dict_results["gradient_updates_per_pass_count"] = self.gradient_updates_per_pass_count
         dict_results["is_early_stopping"] = self.is_early_stopping
-        dict_results["federated_test_score"] = self.federated_test_score
+        dict_results["federated_test_score"] = self.federated_test_result[0]
+        dict_results["federated_test_nb_epochs_done"] = self.federated_test_result[1]
```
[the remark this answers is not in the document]
RomainGoussault: As said before, one option would be to add a MPL attribute to the scenario class

multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+
+        # Evaluate trained model
+        model_evaluation = model.evaluate(x_test, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        logger.info(f"   Model evaluation on test data: "
+                    f"{list(zip(model.metrics_names, ['%.3f' % elem for elem in model_evaluation]))}")
+
+        model_eval_score = model_evaluation[1]  # 0 is for the loss
+
+        # Return model score on test data
+        nb_epochs_done = (es.stopped_epoch + 1) if es.stopped_epoch != 0 else self.epoch_count
```
RomainGoussault, line 120: I don't understand this line

multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+            sequentially_trained_model = utils.generate_new_cnn_model()
+
+        # Train model (iterate for each epoch and mini-batch)
+        for epoch_index in range(epoch_count):
+
+            self.epoch_index = epoch_index
+
+            # Clear Keras' old models (except if the approach is sequential and the model has to persist across epochs)
+            if self.learning_approach != 'seq':
+                clear_session()
```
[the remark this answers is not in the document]
RomainGoussault: Ok I tried and it crashed ... I don't know why though.
It is a problem since I am getting OOM error for `seq ` experiments.

multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+        logger.info(f"   Model metrics names: {model_to_evaluate.metrics_names}")
+        logger.info(f"   Model metrics values: {['%.3f' % elem for elem in model_evaluation]}")
+        test_score = model_evaluation[1]  # 0 is for the loss
+
+        # Plot training history
+        if self.is_save_fig:
+
+            # Save data
+            history_data = {}
+            history_data["score_matrix"] = score_matrix
```
RomainGoussault, line 217: in `update_iterative_results()` this is the attribute `self.score_matrix` which is updated.
I think we should remove the local variables `score_matrix`, `score_matrix_extented `and only use the attributes

multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+            return self.compute_test_score_for_single_partner(partners_list[0])
+
+        # Else, continue onto a federated learning procedure
+        partners_list = sorted(partners_list, key=operator.attrgetter("id"))
+        logger.info(
+            f"## Training and evaluating model on partners with ids: {['#' + str(p.id) for p in partners_list]}")
+
+        # Initialize variables
+        score_matrix, score_matrix_extended = self.score_matrix, self.score_matrix_extended
+        global_val_acc, global_val_loss = self.global_val_acc, self.global_val_loss
```
RomainGoussault, line 147: Potentially to be removed see comment below

multi_partner_learning.py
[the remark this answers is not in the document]
RomainGoussault: Maybe we just have to recompile the model after `clear_session()`

multi_partner_learning.py
bowni: Via the callback the EarlyStopping provides `stopped_epoch`. But the way it is implemented, if early stopping wasn't trigger, this returns 0. So this line is a ternary operator for choosing between `stopped_epoch` and `epoch_count`.

multi_partner_learning.py
bowni: Yes, I am on it (I started working on basics that will be needed for the federated step-by-step increments, and it includes better score matrixes).

multi_partner_learning.py
bowni: This is a kind of "model deep copy" function. I use it in the seq and seqavg.

multi_partner_learning.py
```diff
@@ -0,0 +1,532 @@
...
+        for weights_for_layer in weights_per_layer:
+            avg_weights_for_layer = np.average(
+                np.array(weights_for_layer), axis=0, weights=self.aggregation_weights
+            )
+            new_weights.append(list(avg_weights_for_layer))
+
+        return new_weights
+
+    @staticmethod
+    def build_aggregated_model(new_weights):
```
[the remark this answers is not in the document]
bowni: Ok, to be renamed

multi_partner_learning.py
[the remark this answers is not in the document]
bowni: Hmm, I tried and failed. I let this optimization to you, this is not my strong suit.

multi_partner_learning.py
RomainGoussault: Which line ? I could not find where it was called

multi_partner_learning.py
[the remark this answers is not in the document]
RomainGoussault: Ok https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/142

multi_partner_learning.py
bowni: Ah yeah you are right I didn't need it anymore once I started storing weights instead of models. You might need it though when you'll address the `clear_session()` #142 issue (potentially to be transformed to a `build_from_weights` function).

bowni: Ahhh, in fact this is exactly the role of the function you commented above: `build_aggregated_model()` to be renamed to `build_from_weights()`. Ok so I'll simplify these

bowni:
Ok @RomainGoussault, done. I let you re-review! Also, @natct10, more than happy to get yours!

multi_partner_learning.py
```diff
@@ -67,9 +61,11 @@ def __init__(self,
         self.aggregation_weights = []
         self.models_weights_list = [None] * self.partners_count
         self.local_score_list = [None] * self.partners_count
-        self.score_matrix_minibatch_end = np.zeros(shape=(self.epoch_count, self.minibatch_count, self.partners_count))
-        self.score_matrix_minibatch_start = np.zeros(shape=(self.epoch_count, self.minibatch_count))
-        self.global_val_acc, self.global_val_loss = [], []
+        self.score_matrix_per_partner = np.zeros(shape=(self.epoch_count, self.minibatch_count, self.partners_count))
+        self.score_matrix_collective_models = np.zeros(shape=(self.epoch_count, self.minibatch_count + 1))
```
RomainGoussault, line 65: Why `+ 1` ?

scenario.py
```diff
@@ -36,7 +38,8 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1):
         self.y_test = y_test
 
         # Performance of the model trained in a distributed way on all partners
-        self.federated_test_result = []
+        self.mpl_test_score = None
+        self.mpl_nb_epochs_done = int
```
RomainGoussault, line 42: This is just a copy of `self.mpl.test_score` ? If this is the case, I would remove those two attributes.

multi_partner_learning.py
```diff
@@ -0,0 +1,522 @@
...
+        self.minibatched_x_train = [None] * self.partners_count
+        self.minibatched_y_train = [None] * self.partners_count
+        self.aggregation_weights = []
+        self.models_weights_list = [None] * self.partners_count
+        self.local_score_list = [None] * self.partners_count
+        self.score_matrix_per_partner = np.zeros(shape=(self.epoch_count, self.minibatch_count, self.partners_count))
+        self.score_matrix_collective_models = np.zeros(shape=(self.epoch_count, self.minibatch_count + 1))
+        self.loss_collective_models = []
+        self.test_score = None
+        self.nb_epochs_done = int
```
RomainGoussault, line 68: We should aim to simply the epochs attribute. We have three right now:

- `nb_epochs_done `
- `epoch_index `
- `epoch_count `

Maybe we should rename `epoch_count `to something like `max_epoch_count `and remove `nb_epochs_done `

RomainGoussault: And `nb_epochs_done`= `epoch_index `?

multi_partner_learning.py
```diff
@@ -0,0 +1,522 @@
...
+
+            # At the end of each epoch, evaluate the model for early stopping on a global validation set
+            # Even in the 'seq' approach we aggregate here...
+            # ... to minimize the dependence to the order of partners on last minibatch
+            self.prepare_aggregation_weights()
+            model_to_evaluate = self.build_model_from_weights(self.aggregate_model_weights())
+            model_evaluation = model_to_evaluate.evaluate(
+                x_val, y_val, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0,
+            )
+            self.score_matrix_collective_models[epoch_index, minibatch_count] = model_evaluation[1]
```
RomainGoussault, line 178: collective model is in fact the aggregated model ? 
Maybe rename to `test_score_aggregated_model` ?
And for the loss: `loss_aggregated_model` ?

multi_partner_learning.py
```diff
@@ -0,0 +1,522 @@
...
+
+    def update_iterative_results(self, partner_index, fit_history):
+        """Update the results arrays with results from the collaboration round"""
+
+        validation_score = fit_history.history["val_accuracy"][0]
+
+        self.local_score_list[partner_index] = validation_score  # TO DO check if coherent
+
+        # At the end of each mini-batch, for each partner, populate the score matrix per partner
+        self.score_matrix_per_partner[self.epoch_index, self.minibatch_index, partner_index] = validation_score
```
RomainGoussault, line 501: validation_score_per_partner ?

multi_partner_learning.py
bowni: We save it at beginning of each mini-batch, plus at the end of the last of each epoch.

scenario.py
bowni: Yes, that seems better you are right!

multi_partner_learning.py
bowni: In multi-partner setting yes. In single-partner setting, this is derived from an attribute of the `EarlyStopping()` instance.

multi_partner_learning.py
bowni: Yes but no: it is for the aggregated models in `fedavg` and `seqavg` approaches. In the `seq` approach it is (almost always) the model passed sequentially from mini-batch to mini-batch and partner to partner.

scenario.py
bowni: Done

main.py
```diff
@@ -133,14 +133,11 @@ def run_scenario(current_scenario):
...
 
     # Train and eval on all partners according to scenario
-    start = timer()
     current_scenario.mpl = multi_partner_learning.init_multipartnerlearning_from_scenario(
         current_scenario,
         is_save_fig=True,
     )
-    current_scenario.compute_mpl()
-    end = timer()
-    current_scenario.federated_computation_time_sec = end - start
```
RomainGoussault: I think we should revert this change. The timer created inside the `compute_mpl()` method will conflict with the timers created when doing contributivity estimation, for instance here --> https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/8e5a242d150bf7e73853d9377984295047d59474/contributivity.py#L149

bowni: Hmm, interesting. Don't you think the respective variable scopes isolate `start` and `end`?

bowni: Ah no, I get what you mean. The result stored as a `Scenario` attribute will be overwritten each time `compute_mpl()` is called within a contributivity approach.

bowni: Actually I think it won't: the result stored as a `Scenario` attribute is from the `MultiPartnerLearning` object that is itself a `Scenario` attribute. When a contributivity approach is called, and method `not_twice_characteristic()`, a dedicated `MultiPartnerLearning` object is instantiated. I don't see where there would be a conflict. Do you still do @RomainGoussault ?

RomainGoussault: Actually I think you are right. I thought timer() was doing some complex stuff, but it's not the case.

multi_partner_learning.py
natct10: Looks cool!

Issue #140, opened 2020-05-19
Investigate why we don't see any catastrophic forgetting 


RomainGoussault:
We should have come across [Catastrophic forgetting](https://ai.stackexchange.com/a/13293
) when doing experiment with the [seq ](https://github.com/SubstraFoundation/distributed-learning-contributivity#configuration-of-the-collaborative-and-distributed-learning)approach.

We did not, and I think we should aim to provoke this Catastrophic forgetting, to be sure our distributed learning code is not flawed. I think we should:

- Have a real sequential approach (without any averaging)
- Run simulation on stratified datasets with different minibatch sizes.

cc @bowni @Thomas-Galtier

bowni:
I tried a few scenarios, indeed it seems difficult to observe. An hypothesis: because the validation dataset used in the `fit()` is global and representative of the whole data space; with a 'local' validation dataset corresponding to the data space of the local partner, we would overfit it at each round and observe the 'forgetting'.

What do you think @RomainGoussault @Thomas-Galtier ?

RomainGoussault:
Below some results with the `seq-pure` approach:
![image](https://user-images.githubusercontent.com/5354301/83856863-7c065700-a71a-11ea-9f7a-b7adb4e6badc.png)

The trends looks OK but I am unsure if we should see such good results with the pure-seq approch with one minibatch.


Two things we can try to investigate further:

- Look at the loss for the pure-seq approch with one minibatch.
- Implement a new file that learns on MNIST (without any multi partner learning) but with a stratified dataste and check if we see any castastrophic forgetting.

RomainGoussault:
Also we should try with a more stratified scenario with 10 partner, each one with only one digit.

RomainGoussault:
We don't see much in the loss --> https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/486f58b91fd946ac1cf766fa1a859a9a3adc9796/notebooks/mpl_with_losses.ipynb

RomainGoussault:
I launched the following new experiment:
```
   amounts_per_partner: 
     - [0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
   samples_split_option: 
[...]

Issue #146, opened 2020-05-26
Create "portfolio" of past experiments


RomainGoussault:
Possibilities:

- notebook (with or without binder)
- wiki
- ?

natct10:
I was also thinking about https://osf.io/ to publish some results we would like to publish with a DOI (citation reference) and in a metedata harvestable format. What do you think about it?

natct10:
> I was also thinking about https://osf.io/ to publish some results with a DOI (citation reference) and in a metedata harvestable format. What do you think about it?

Osf is fire! cc @bowni

PR #152, opened 2020-06-10
Portfolio

RomainGoussault:
I added an experiments that compare different distributed learning approches for MNIST and CIFAR10.

There are lots files added. I propose that we reduce the number of files created by each experiment https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/174.

You should probably only look at this notebook; https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/portfolio/saved_experiments/mnist_cifar10_distributed_learning/mnist_cifar10_distributed_learning.ipynb


bowni:
@RomainGoussault aren't we good on that one with the experiment you showed us on Tuesday? I'd say this closes #140 and #146

PR #153, opened 2020-06-11
Implement the DVRL approach

bowni:


bowni:
@Thomas-Galtier do you confirm we drop this one and focus on PVRL?

Thomas-Galtier:
yep

bowni:
Dropping it! cc @Thomas-Galtier @arthurPignet @RomainGoussault

PR #155, opened 2020-06-11
Validate scenario before running them

### LINKED ISSUE #132: All scenario should be "validated" before starting any computation
When running multiple scenarios, the scenarios are run one by one. If one fails, we will only know when it is running --> We should know at the very start of the program.

RomainGoussault:
Fixes #132

### DIFF
--- a/main.py
+++ b/main.py
@@ -121,15 +121,24 @@ def move_log_file_to_experiment_folder(logger_id, experiment_path, filename, lev
     logger.add(new_log_path, level=level)
 
 
-
 def validate_scenario_list(scenario_params_list, experiment_path):
+    """Instanciate every sceneario withouth running it to check if
+    every scenario is correctly specified"""
 
     logger.debug("Starting to validate scenarios")
+
     for scenario_id, scenario_params in enumerate(scenario_params_list):
 
         logger.debug(f"Validation scenario {scenario_id + 1}/{len(scenario_params_list)}")
-        scenario.Scenario(scenario_params, experiment_path, is_logged_scenario=False)
-    
+
+        current_scenario = scenario.Scenario(scenario_params, experiment_path, is_logging_enabled=False)
+        current_scenario.instantiate_scenario_partners()
+
+        if isinstance(current_scenario.samples_split_option, list):
+            current_scenario.split_data_advanced()
+        else:
+            current_scenario.split_data(is_logging_enabled=False)        
+
     logger.debug("All scenario have been validated")
 
 
--- a/main.py
+++ b/main.py
@@ -42,6 +42,8 @@ def main():
         experiment_path = config["experiment_path"]
         n_repeats = config["n_repeats"]
 
+        validate_scenario_list(scenario_params_list, experiment_path)
+
         for scenario_id, scenario_params in enumerate(scenario_params_list):
             logger.info(f"Scenario {scenario_id+1}/{len(scenario_params_list)}: {scenario_params}")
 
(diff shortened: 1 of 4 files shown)


main.py
```diff
@@ -119,6 +121,27 @@ def move_log_file_to_experiment_folder(logger_id, experiment_path, filename, lev
     logger.add(new_log_path, level=level)
 
 
+def validate_scenario_list(scenario_params_list, experiment_path):
```
bowni, line 124: Couldn't we save the `Scenarios` instantiated, with their data split, so that we don't re-instantiate them and re-split data?

main.py
```diff
@@ -119,6 +121,27 @@ def move_log_file_to_experiment_folder(logger_id, experiment_path, filename, lev
...
+    """Instanciate every sceneario withouth running it to check if
+    every scenario is correctly specified"""
```
natct10, line 126: *without

celinejacques: *scenario

requirements.txt
```diff
@@ -7,4 +7,4 @@ pandas==0.25.3
 seaborn==0.10.0
 loguru==0.4.1
 tensorflow==2.1.0
-PyYAML~=5.3.1
\ No newline at end of file
+ruamel.yaml~=0.16.10
```
natct10, line 10: Oh cool!

utils.py
```diff
@@ -65,9 +65,11 @@ def load_cfg(yaml_filepath):
         cfg : dict
     """
     logger.info("Loading experiment yaml file")
-    # Read YAML experiment definition file
+
+    yaml=YAML(typ='safe')
     with open(yaml_filepath, "r") as stream:
-        cfg = yaml.load(stream, Loader=yaml.FullLoader)
+        # This will fail if there are duplicated keys in the YAML file
```
natct10, line 71: So, why not add a try/except?

main.py
RomainGoussault: We could, but I think it would be a bit messy : some scenerio stuff would be done in the `validate_scenario_list` method (like `split_data()` for instance and other (like `preprocess_scenarios_data`) would be done in the `run_scenario `method

main.py
RomainGoussault: Done

utils.py
RomainGoussault: We (I) want it to fail quickly if there is a duplicated key because the only solution is for the user to modify the config file.

main.py
bowni: What is the most Pythonic:

- do the scenario instanciation and the split data twice
- or do it once, and then do following operations outside of the validation loop

?

My intuition would be the second one. Maybe by just renaming your method to something like `instantiate_and_validate...` that would look clearer. No?

natct10: Hey, I'm not sure about the most efficient way to organize it, maybe we could use `timeit`?

RomainGoussault: We should not save the scenarios instanciated during the validation because, we would anyway need to re-create them for the other random states.

RomainGoussault: And there is no need to explicitely call `del current_scenario`, as all the variables will be freed when the function will exit.

bowni: Ok then

RomainGoussault:
> @RomainGoussault so if it fails the user will just see the assertion error of the scenario that failed. Couldn't we catch the assertion error and, in addition to displaying it, display a message like "Hey, it seems at least one of the scenarios derived from your config files didn't go through validation. So nothing will be launched. Have fun then"?

@bowni We could catch the assertion error, log a message and re-throw another error. But actually we already log "Validation scenario X/30" at each scenario so I think it's pretty explanatory that the assert error was done during the scenario validation ?
Another think is to have nice assert messages and I think it's not the case right now.

PR #157, opened 2020-06-11
Log result.csv path

natct10:
This is a micro submission to add the full path of results.csv in the log. 
Shall I go further?


natct10:
Ow I was based on another branch. I'll reopen another one :/

PR #158, opened 2020-06-11
Log result.csv path

natct10:
This is a micro submission to add the full path of results.csv in the log.
Shall I go further?

### DIFF
--- a/main.py
+++ b/main.py
@@ -9,19 +9,21 @@ A script for:
 from timeit import default_timer as timer
 import matplotlib.pyplot as plt
 import numpy as np
-import utils
 from loguru import logger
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -84,7 +84,7 @@ def main():
 
                 with open(experiment_path / "results.csv", "a") as f:
                     df_results.to_csv(f, header=f.tell() == 0, index=False)
-                    logger.info("Results saved")
+                    logger.info(f"Results saved to {str(experiment_path)}/results.csv")
```
RomainGoussault, line 87: Maybe we should not log the whole filepath but only the parts relative to the git folder ?

PR #159, opened 2020-06-11
add model stacking

Thomas-Galtier:
model stacking is a method for model ensembling


multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
...
+                batch_size=partner.batch_size,
+                epochs=self.epoch_count,
+                verbose=0,
+                validation_data=self.val_data,
+                callbacks=cb,
+            )
+            #store the model
+            partner.model=model
+            
+            self.nb_epochs_done = (es.stopped_epoch + 1) if es.stopped_epoch != 0 else self.epoch_count
```
celinejacques, line 109: I assume the number of epochs can be different from a partner to another. Moreover, we maybe should differentiate the number of epochs of single training and the number used to train the meta_model for model stacking and keep all to analyze results of experiments.

main.py
```diff
@@ -137,7 +137,10 @@ def run_scenario(current_scenario):
         current_scenario,
         is_save_data=True,
     )
-    current_scenario.mpl.compute_test_score()
+    if current_scenario.multi_partner_learning_approach=="stacking":
```
natct10, line 140: Need space here: `approach == "stacking"`

multi_partner_learning.py
```diff
@@ -33,7 +37,7 @@ def __init__(self,
                  single_partner_test_mode="global",
                  is_early_stopping=True,
                  is_save_data=False,
-                 save_folder="",
```
natct10, line 36: You can have a comma here, so it minimizes the diff when you add another param

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
         self.is_save_data = is_save_data
         self.save_folder = save_folder
         self.learning_computation_time = None
+        
 
         logger.debug("MultiPartnerLearning object instantiated.")
 
+    def train_single_model(self,partner):
+        #first check if the model was already trained
+        if not partner.model==None:
```
natct10, line 83: This might be simplified!

multi_partner_learning.py
```diff
@@ -9,6 +9,10 @@
 import keras
 from keras.backend.tensorflow_backend import clear_session
 from keras.callbacks import EarlyStopping
+from keras.models import Model
+from keras.layers import Input
+from keras.layers import Dense
```
bowni, line 14: I think you can do `from keras.layers import Input, Dense`

multi_partner_learning.py
```diff
@@ -54,7 +58,7 @@ def __init__(self,
         self.epoch_index = 0
         self.minibatch_count = minibatch_count
         self.minibatch_index = 0
-        self.is_early_stopping = is_early_stopping
+        self.is_early_stopping = is_early_stopping
```
bowni, line 61: Unnecessary whitespace at the end of the line?

multi_partner_learning.py
bowni: Probably `if partner.model:` ?

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
         self.is_save_data = is_save_data
         self.save_folder = save_folder
         self.learning_computation_time = None
+        
 
         logger.debug("MultiPartnerLearning object instantiated.")
 
+    def train_single_model(self,partner):
+        #first check if the model was already trained
```
bowni, line 82:
```suggestion
        # First, check if the model was already trained
```

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
...
+                partner.x_train,
+                partner.y_train,
+                batch_size=partner.batch_size,
+                epochs=self.epoch_count,
+                verbose=0,
+                validation_data=self.val_data,
+                callbacks=cb,
+            )
+            #store the model
+            partner.model=model
```
bowni, line 107:
```suggestion
            )
            
            # Store the model
            partner.model = model
```

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
...
+        model_evaluation = meta_model.evaluate(X, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        logger.info(f"   Model evaluation on test data: "
+                    f"{list(zip(meta_model.metrics_names, ['%.3f' % elem for elem in model_evaluation]))}")   
+        self.test_score = model_evaluation[1]  # 0 is for the loss
+        end = timer()
+        self.learning_computation_time = end - start
+            
     def compute_test_score_for_single_partner(self, partner):
-        """Return the score on test data of a model trained on a single partner"""
+        """compute test the score on test data of a model trained on a single partner"""
```
bowni, line 171:
```suggestion
        """ Compute the score on test data of a model trained on a single partner"""
```

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
...
-            partner.y_train,
-            batch_size=partner.batch_size,
-            epochs=self.epoch_count,
-            verbose=0,
-            validation_data=self.val_data,
-            callbacks=cb,
-        )
-
+        model=self.train_single_model(partner)
         # Reference a testset according to the scenario configuration
```
bowni, line 177:
```suggestion
        model = self.train_single_model(partner)
        
        # Reference a testset according to the scenario configuration
```

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
...
+        start = timer()
+        x_val, y_val = self.val_data
+        x_test, y_test = self.test_data
+        meta_model=self.make_stacked_meta_model(meta_model_hidden_dim=constants.NUM_CLASSES)
+        logger.info(f"## Meta-model compiled.")
+        # prepare input data
+        X = [x_val for _ in range(len(meta_model.input))]
+        # fit model
+        logger.info(f"## Fitting the meta-model...")
+        meta_model.fit(X, y_val, epochs=100, verbose=0)
```
bowni, line 157:
```suggestion
    def compute_test_score_with_stacking(self):
    
        # First, if only one partner, fall back to dedicated single partner function
        if self.partners_count == 1:
            return self.compute_test_score_for_single_partner(self.partners_list[0])
        
        logger.info(f"## Training and evaluating stacked model on partners with ids: {['#' + str(p.id) for p in self.partners_list]}")
        start = timer()
        x_val, y_val = self.val_data
        x_test, y_test = self.test_data
        meta_model = self.make_stacked_meta_model(meta_model_hidden_dim=constants.NUM_CLASSES)
        logger.debug(f"## Meta-model compiled.")
        
        # Prepare input data
        X = [x_val for _ in range(len(meta_model.input))]
        
        # Fit model
        logger.debug(f"## Fitting the meta-model...")
        meta_model.fit(X, y_val, epochs=100, verbose=0)
```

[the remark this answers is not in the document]
bowni: Yes that would be optimal, at least to parameterize the epochs.

multi_partner_learning.py
```diff
@@ -70,36 +74,106 @@ def __init__(self,
...
+        # Evaluate trained model
+        logger.info(f"## Evaluating the meta-model...")
+        ## prepare test data
+        X = [x_test for _ in range(len(meta_model.input))]
+        model_evaluation = meta_model.evaluate(X, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        logger.info(f"   Model evaluation on test data: "
+                    f"{list(zip(meta_model.metrics_names, ['%.3f' % elem for elem in model_evaluation]))}")   
+        self.test_score = model_evaluation[1]  # 0 is for the loss
+        end = timer()
+        self.learning_computation_time = end - start
```
bowni, line 168:
```suggestion
        # Evaluate trained model
        logger.debug(f"## Evaluating the meta-model...")
        
        # Prepare test data
        X = [x_test for _ in range(len(meta_model.input))]
        model_evaluation = meta_model.evaluate(X, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
        logger.debug(f"   Model evaluation on test data: "
                     f"{list(zip(meta_model.metrics_names, ['%.3f' % elem for elem in model_evaluation]))}")   
        self.test_score = model_evaluation[1]  # 0 is for the loss
        end = timer()
        self.learning_computation_time = end - start
```

Thomas-Galtier:
@bowni 
To understand stacking I first read the section on stacking from [this article](https://towardsdatascience.com/ensemble-methods-bagging-boosting-and-stacking-c9214a10a205i )
I also watched this [video](https://www.youtube.com/watch?v=sBrQnqwMpvA&t=319s) which helped me understand how it works when one aggregate the outputs of several models with a meta model.

Finally I used the code from 
[this kaggle tuturial](https://machinelearningmastery.com/stacking-ensemble-for-deep-learning-neural-networks/)

multi_partner_learning.py
```diff
@@ -73,54 +82,151 @@ def __init__(self,
...
+        ensemble_outputs = [model.output for model in list_of_models]
+        merge = concatenate(ensemble_outputs)
+        hidden = Dense(meta_model_hidden_dim, activation="relu")(merge)
+        output = Dense(constants.NUM_CLASSES, activation="softmax")(hidden)
+        meta_model = Model(inputs=ensemble_visible, outputs=output)
+        # compile
+        meta_model.compile(
+            loss="categorical_crossentropy", optimizer="adam", metrics=["accuracy"]
+        )
+        return meta_model
```
bowni, line 151:
```suggestion
    def make_stacked_meta_model(self, meta_model_hidden_dim=10):
    
        logger.info("## Model ensembling approach - Creating the meta-model")
        
        # Create a list with an independently trained model for each partner
        list_of_models = []
        for partner in self.partners_list:
            list_of_models.append(self.train_single_model(partner))

        # Update all layers in all models to not be trainable
        for i in range(len(list_of_models)):
            model = list_of_models[i]
            for layer in model.layers:
                layer.trainable = False  # Make it not trainable
                layer.name = f"ensemble_{i+1}_{layer.name}"  # Rename it to avoid 'unique layer name' issue
                
        # Define multi-headed input
        ensemble_visible = [model.input for model in list_of_models]
        
        # Concatenate merge output from each model
        ensemble_outputs = [model.output for model in list_of_models]
        merge = concatenate(ensemble_outputs)
        hidden = Dense(meta_model_hidden_dim, activation="relu")(merge)
        output = Dense(constants.NUM_CLASSES, activation="softmax")(hidden)
        meta_model = Model(inputs=ensemble_visible, outputs=output)
        
        # Compile the meta-model
        meta_model.compile(
            loss="categorical_crossentropy", optimizer="adam", metrics=["accuracy"]
        )
        
        return meta_model
```

multi_partner_learning.py
```diff
@@ -70,54 +80,155 @@ def __init__(self,
...
+
+    def compute_test_score_with_stacking(self):
+        
+        # First, if only one partner, fall back to dedicated single partner function
+        if self.partners_count == 1:
+            return self.compute_test_score_for_single_partner(self.partners_list[0])
+
+        logger.info(
+            f"## Training and evaluating stacked modelon partners with ids: {['#' + str(p.id) for p in self.partners_list]}"
```
RomainGoussault, line 162: typo: modelon

multi_partner_learning.py
```diff
@@ -70,54 +80,155 @@ def __init__(self,
...
+        start = timer()
+        x_val, y_val = self.val_data
+        x_test, y_test = self.test_data
+        meta_model = self.make_stacked_meta_model(
+            meta_model_hidden_dim=self.num_classes 
+        )
+        logger.info("## Meta-model compiled.")
+        
+        # prepare input data
+        X = [x_val for _ in range(len(meta_model.input))]
```
RomainGoussault, line 173: Something like x_val_expanded ?

multi_partner_learning.py
```diff
@@ -70,54 +80,155 @@ def __init__(self,
 
         logger.debug("MultiPartnerLearning object instantiated.")
 
+    def train_single_model(self, partner):
+        # first check if the model was not already trained
+        if partner.model == None:
```
natct10, line 85: Ici tu pourrais faire `if not partner.model:`

main.py
```diff
@@ -163,7 +163,12 @@ def run_scenario(current_scenario):
         current_scenario,
         is_save_data=True,
     )
-    current_scenario.mpl.compute_test_score()
+    if current_scenario.multi_partner_learning_approach == "stacking":
+        current_scenario.mpl.compute_test_score_with_stacking()
+        for partner in current_scenario.partner_list:
+            partner.model=None #clear the models (in order to compute well the time of the method used after the computation of the federated score)
```
natct10, line 169: Need space: `partner.model = None`

multi_partner_learning.py
```diff
@@ -52,15 +57,20 @@ def __init__(self,
         self.minibatch_count = minibatch_count
         self.minibatch_index = 0
         self.is_early_stopping = is_early_stopping
-
+        self.epoch_count_for_meta_model=epoch_count_for_meta_model
```
natct10, line 60: Need space :D

multi_partner_learning.py
```diff
@@ -504,25 +694,33 @@ def collaborative_round_fit(model_to_fit, train_data, val_data, batch_size):
...
 
     def update_iterative_results(self, partner_index, fit_history):
         """Update the results arrays with results from the collaboration round"""
 
         validation_score = fit_history.history["val_accuracy"][0]
 
-        self.scores_last_learning_round[partner_index] = validation_score  # TO DO check if coherent
+        self.scores_last_learning_round[
+            partner_index
+        ] = validation_score  # TO DO check if coherent
```
natct10, line 718: Is it possible the rearrange this definition as `validation_score = (...)`?

Thomas-Galtier: What do you mean by rearranging?

bowni:
@Thomas-Galtier in the above thread could you "_Resolve conversation_" when the comment is taken care of or obsolete? It will help getting a better visibility on what is left to review / update!

multi_partner_learning.py
natct10: I might have glitched on this one, my bad!

bowni:
- [ ] Training on train data instead of val dat
- [ ]  Investigate log of first experiments, duration of computation, number of epochs...

README.md
```diff
@@ -194,6 +194,10 @@ Define the multi-partner learning approach, among the following as described by
     
     ![Schema seqavg](img/collaborative_rounds_seqavg.png)
     
+ - `'stacking'`: stands for model stacking
+    
+    ![Schema stacking](img/model_stacking.png)
```
bowni, line 199: In the schema, could you @Thomas-Galtier figure that the original inputs are also used as train inputs of the meta-model?

main.py
```diff
@@ -171,6 +171,12 @@ def run_scenario(current_scenario):
         is_save_data=True,
     )
     current_scenario.mpl.compute_test_score()
+    # if satcking clear the partners' model (in order to compute well the time of the methods used after the computation of the federated score)
+    if current_scenario.multi_partner_learning_approach == "stacking": 
+        for partner in current_scenario.partners_list:
+            partner.model=None
```
bowni, line 177: There is no `model` attribute to the `Partner` object. Do you target `model_weights` here?

bowni: I have the feeling that this action would be better located at the end of `compute_test_score()`, don't you think?
(Rationale behind is to try to keep only high level functions in `main.py`)

multi_partner_learning.py
```diff
@@ -124,28 +193,39 @@ def compute_test_score(self):
...
         partners_list = sorted(partners_list, key=operator.attrgetter("id"))
         logger.info(
-            f"## Training and evaluating model on partners with ids: {['#' + str(p.id) for p in partners_list]}")
+            f"## Training and evaluating model on partners with ids: {['#' + str(p.id) for p in partners_list]}"
+        )
 
         # Initialize variables
-        model_to_evaluate, sequentially_trained_model = None, None
-        if self.learning_approach in ['seq-pure', 'seq-with-final-agg']:
+        model_to_evaluate, sequentially_trained_model, stacking_model = None, None, None
```
bowni, line 214: Here we could keep only one of the two (`sequentially_trained_mode` and `stacking_model`), as only one of them is used downstream; it would require giving it a more generic name, in substance "the model that we use across multiple epochs"). But would that really be better, I am not too sure for readibility. What do you think @Thomas-Galtier @RomainGoussault @natct10 ?

multi_partner_learning.py
```diff
@@ -156,52 +236,75 @@ def compute_test_score(self):
...
+                    self.compute_collaborative_round_sequential(
+                        sequentially_trained_model
+                    )
 
-                elif self.learning_approach == 'seqavg':
+                elif self.learning_approach == "seqavg":
                     self.compute_collaborative_round_seqavg()
+                
+                elif self.learning_approach == "stacking":
+                    self.compute_collaborative_round_stacking(stacking_model)
```
bowni, line 251: @Thomas-Galtier there is something I don't get 😃 Why do you train the meta-model with this custom `epochs x mini-batches` approach, why not just 1 single `.fit()`?

Thomas-Galtier: Well training the meta-model is like training the main model in other approaches ( sequential or fedavg ). The `epoch x minibatch` approach should  improve the quality of the fit of the model, especially in stratified cases.

bowni: Well, in this case it seems to me that computation-wise it doesn't present an advantage versus the other approaches.

My understanding before was that:

- we do an independent training on each partner's data; this costs 1 'full-scale' computation
- then for any coalition of partners that we want to assess, we can assemble their models in a meta-model; this costs not much
- this way, a Shapley-based approach would be less costly, as it wouldn't require 'full-scale' training for each coalition to assess

But eventually if for a given coalition of partners we have to do a 'full scale' training on their data, this rationale doesn't hold. Sorry if it was clear for some time, I just figured this out now.

Some questions @Thomas-Galtier :

- So in a sense, we could see the approach you implemented as starting with a pre-trained model? The assembly of the individual models create a kind of pre-trained model, which we train on all data?
- Or, in a Shapley-like approach where we need to assess different coalition of partners, could we do the full-scale training only for the larger coalition, and derive the other ones from what we learned here (without the need to re-train the meta-model)?

Thomas-Galtier: * My hope was that the full-scale training on the meta-model would be quicker than  a full-scale training with the main model, because the meta model is lighter. But It does not seems to be the case. I  guess   that tensorflow still computes the gradients  of all the architecture of the stacked model ( for each  partner's  model  and for the meta-model), and then only apply the gradient  of the meta model part (as the rest of the architecture is set to not trainable variables). Thus it computes a lot of gradients that it does not uses in the end, that's why it is so slow to train the meta model...   I may have an idea to avoid theses extra computations : Instead of training the stacked model we can train only the meta model by setting its input as the ouputs of each partner model.

* No the approach cannot be seen as starting with a pretrained model, as the architecture of the stacked model (N models stacked in parallel + a meta model to aggregate their outputs) is different from the architecture of a single partner model .

* I don't know how  we can use the stacked model on the larger coalition to infer what would be a stacked model on a smaller coalition as the two stacked model on each coalition do not have the same architecture.

main.py
natct10: typo in the comment: "satcking" => stacking"

PR #160, opened 2020-06-12
Open the way to the implementation of unit tests and config for Travis CI

bowni:


natct10:
@bowni It seems you can't review as you are involved in the PR

PR #166, opened 2020-06-17
Iss144 initialize experiment with an existing model

### LINKED ISSUE #144: Initialize model with model from a previous coalition?
This could "save" a few epochs.

celinejacques:
closes #144

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
        gradient_updates_per_pass_count:
          - 8
        dataset_proportion:
-	 - 1
+	     - 1
      - dataset_name:
          - 'mnist'
        partners_count:
@@ -140,7 +140,7 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
        gradient_updates_per_pass_count:
          - 8
        dataset_proportion:
-	 - 1
+	     - 1
     ```
 
    Under `scenario_params_list`, enter a list of sets of scenario(s). Each set starts with ` - dataset_name:` and must have only one `partners_count` value. The length of `amount_per_partners`, `corrupted_datasets` (and `samples_split_option` when the advanced definition is used) must match the `partner_counts` value. If for a given parameter multiple values are specified, e.g. like for `agregation_weighting` in the first scenario set of the above example, all possible combinations of parameters will be assembled as separate scenarios and run.
--- a/README.md
+++ b/README.md
@@ -93,9 +93,9 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
     scenario_params_list:
      - dataset_name:
        'mnist':
-        - 'init'
+        - 'random_initialization'
        'cifar10':
-        - 'init'
+        - 'random_initialization'
        partners_count: 
          - 3
        amounts_per_partner:
--- a/README.md
+++ b/README.md
@@ -87,59 +87,61 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
 
 1. Define your mock scenario(s) in `config.yml` by changing the values of the suggested parameters of the 2 example scenarios (you can browse more available parameters in section [Config file parameters](#config-file-parameters) below). For example:
 
-```yaml
-experiment_name: my_custom_experiment
-n_repeats: 5
-scenario_params_list:
-  - dataset_name:
-      - 'mnist'
-      - 'cifar10'
-    partners_count:
-      - 3
-    amounts_per_partner:
-      - [0.4, 0.3, 0.3]
-    samples_split_option:
-      - ['advanced', [[7, 'shared'], [6, 'shared'], [2, 'specific']]]
-    multi_partner_learning_approach:
-      - 'fedavg'
-    aggregation_weighting:
-      - 'data_volume'
-      - 'uniform'
-    epoch_count:
-      - 38
-    methods:
-      - ["Shapley values", "Independent scores", "TMCS"]
-    minibatch_count:
-      - 20
-    gradient_updates_per_pass_count:
-      - 8
-    dataset_proportion:
-      - 1
-  - dataset_name:
-      - 'mnist'
-    partners_count:
-      - 2
-    amounts_per_partner:
-      - [0.5, 0.5]
-    samples_split_option:
-      - ['basic', 'random']
-      - ['basic', 'stratified']
-    multi_partner_learning_approach:
-      - 'fedavg'
-    aggregation_weighting:
-      - 'data_volume'
-      - 'uniform'
-    epoch_count:
-      - 38
-    methods:
-      - ["Shapley values", "SMCS", "IS_lin_S", "IS_reg_S"]
-    minibatch_count:
-      - 20
-    gradient_updates_per_pass_count:
-      - 8
-    dataset_proportion:
-      - 1
-```
+    ```yaml
+    experiment_name: my_custom_experiment
+    n_repeats: 5
+    scenario_params_list:
+     - dataset_name:
+       'mnist':
+        - 'init'
+       'cifar10':
+        - 'init'
+       partners_count: 
+         - 3
+       amounts_per_partner:
+         - [0.4, 0.3, 0.3]
+       samples_split_option:
+         - ['advanced', [[7, 'shared'], [6, 'shared'], [2, 'specific']]]
+       multi_partner_learning_approach:
+         - 'fedavg'
+       aggregation_weighting:
+         - 'data_volume'
+         - 'uniform'
+       epoch_count:
+         - 38
+       methods:
+         - ["Shapley values", "Independent scores", "TMCS"]
+       minibatch_count:
+         - 20
+       gradient_updates_per_pass_count:
+         - 8
+       dataset_proportion:
+	 - 1
+     - dataset_name:
+         - 'mnist'
+       partners_count:
+         - 2
+       amounts_per_partner:
+         - [0.5, 0.5]
+       samples_split_option:
+         - ['basic', 'random']
+         - ['basic', 'stratified']
+       multi_partner_learning_approach:
+         - 'fedavg'
+       aggregation_weighting:
+         - 'data_volume'
+         - 'uniform'
+       epoch_count:
+         - 38
+       methods:
+         - ["Shapley values", "SMCS", "IS_lin_S", "IS_reg_S"]
+       minibatch_count:
+         - 20
+       gradient_updates_per_pass_count:
+         - 8
+       dataset_proportion:
+	 - 1
+    ```
 
    Under `scenario_params_list`, enter a list of sets of scenario(s). Each set starts with ` - dataset_name:` and must have only one `partners_count` value. The length of `amount_per_partners`, `corrupted_datasets` (and `samples_split_option` when the advanced definition is used) must match the `partner_counts` value. If for a given parameter multiple values are specified, e.g. like for `agregation_weighting` in the first scenario set of the above example, all possible combinations of parameters will be assembled as separate scenarios and run.
 
@@ -167,9 +169,9 @@ Example: `n_repeats: 2`
 #### Scenario-level parameters
 
 ##### Choice of dataset
-
 `dataset_name`: `'mnist'` (default), `'cifar10'` or `'titanic'`
-MNIST, CIFAR10 and Titanic are currently supported. They come with their associated modules in `/datasets` for loading data, pre-processing inputs, and define a model architecture.
+MNIST, CIFAR10 and Titanic are currently supported. They come with their associated modules in `/datasets` for loading data, pre-processing inputs, and define a model architecture.\
+For each dataset, it is possible to provide a path to model weights learned from a previous coalition. Use `'random_initialization'` if you want a random initialization. 
 
 **Note on validation and test datasets**:
 
--- a/README.md
+++ b/README.md
@@ -171,7 +171,32 @@ Example: `n_repeats: 2`
 ##### Choice of dataset
 `dataset_name`: `'mnist'` (default), `'cifar10'` or `'titanic'`
 MNIST, CIFAR10 and Titanic are currently supported. They come with their associated modules in `/datasets` for loading data, pre-processing inputs, and define a model architecture.\
-For each dataset, it is possible to provide a path to model weights learned from a previous coalition. Use `'random_initialization'` if you want a random initialization. 
+For each dataset, it is possible to provide a path to model weights learned from a previous coalition. Use `'random_initialization'` if you want a random initialization or an empty value as in one of the two following syntaxes:
+
+```
+scenario_params_list:
+ - dataset_name:
+    'mnist':
+    'cifar10':
+```
+
+``` 
+scenario_params_list:
+ - dataset_name:
+    - 'mnist'
+    - 'cifar10'
+```
+
+Please, note that if you want to specify a path to saved weights and launch the same scenario with a random initialization, you must specify it in the list of initialization.
+
+```
+scenario_params_list:
+ - dataset_name:
+    'mnist':
+     - 'random_initialization'
+     - path_to_saved_weights
+    'cifar10':
+```
 
 **Note on validation and test datasets**:
 
--- a/config.yml
+++ b/config.yml
@@ -3,9 +3,9 @@ n_repeats: 5
 scenario_params_list:
  - dataset_name:
     'mnist':
-    - 'init'
+    - 'random_initialization'
     'cifar10':
-    - 'init'
+    - 'random_initialization'
    partners_count:
      - 3
    amounts_per_partner: 
--- a/config.yml
+++ b/config.yml
@@ -3,9 +3,9 @@ n_repeats: 5
 scenario_params_list:
  - dataset_name:
      - 'mnist':
-      -
+      - 'init'
      - 'cifar10':
-      -
+      - 'init'
    partners_count:
      - 3
    amounts_per_partner: 
--- a/config.yml
+++ b/config.yml
@@ -34,7 +34,7 @@ scenario_params_list:
      - 1
  - dataset_name:
     'mnist':
-    - 'init'
+    - 'random_initialization'
    partners_count:
      - 2
    amounts_per_partner: 
--- a/config.yml
+++ b/config.yml
@@ -2,10 +2,10 @@ experiment_name: my_example
 n_repeats: 5
 scenario_params_list:
  - dataset_name:
-     - 'mnist':
-      - 'init'
-     - 'cifar10':
-      - 'init'
+    'mnist':
+    - 'init'
+    'cifar10':
+    - 'init'
    partners_count:
      - 3
    amounts_per_partner: 
@@ -33,8 +33,8 @@ scenario_params_list:
    dataset_proportion:
      - 1
  - dataset_name:
-     - 'mnist':
-      -
+    'mnist':
+    - 'init'
    partners_count:
      - 2
    amounts_per_partner: 
--- a/config.yml
+++ b/config.yml
@@ -1,55 +1,59 @@
 experiment_name: my_example
 n_repeats: 5
 scenario_params_list:
-  - dataset_name:
-    - "mnist"
-    - "cifar10"
-    partners_count:
-      - 3
-    amounts_per_partner:
-      - [0.4, 0.3, 0.3]
-    samples_split_option:
-      - ["basic", "random"]
-      - ["basic", "stratified"]
-      - ["advanced", [[4, "shared"], [6, "shared"], [4, "specific"]]]
-    multi_partner_learning_approach:
-      - "fedavg"
-      - "seq-pure"
-      - "seq-with-final-agg"
-      - "seqavg"
-    aggregation_weighting:
-      - "data_volume"
-      - "uniform"
-    epoch_count:
-      - 38
-    methods:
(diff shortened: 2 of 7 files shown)


scenario.py
```diff
@@ -157,6 +157,12 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1, is_dry_ru
         else:
             self.is_early_stopping = True  # default
 
+        # Model from previous coalition used to initialise model
+        try:
+            self.init_model_from = params["init_model_from"]
+        except:
+            self.init_model_from= ""
+
```
bowni, line 165: @celinejacques is that a Python best practice? If so, it is really cool because I wasn't too fond of our `if ... else ...` for all config parameters. Just tell me, and I can take care of refactoring all those with your `try ... except ...` approach!

bowni:
Wow, this is still draft status but this already looks awesome! A lot of ideas just popped up in my mind after reading your code here @celinejacques :)

scenario.py
celinejacques: I'm not sure that it is a Python best practice and I'm assuming that it depends on each case. Here, I found it usefull. 
@RomainGoussault @natct10 what do you think?

celinejacques:
Thanks @bowni. 
What do you think about the config file and how to specify if a previous model is used to initialize models. It's a bit of headache. I'm sure we can do it in a more elegant way but I've not figured it out yet. If someone has an idea, feel free to share so I can improve it. 

cc @jeromechambost @RomainGoussault @Thomas-Galtier

scenario.py
natct10: I would tend to say that `except` is cool to catch an exception, so here if/else would be sufficient, but I don't have objections

bowni:
> Thanks @bowni.
> What do you think about the config file and how to specify if a previous model is used to initialize models. It's a bit of headache. I'm sure we can do it in a more elegant way but I've not figured it out yet. If someone has an idea, feel free to share so I can improve it.
> 
> cc @jeromechambost @RomainGoussault @Thomas-Galtier @natct10

Good question. So:

- we could see that as an option, either the user wants to activate it to optimize the computation time of its experiment, or not use it  for a really 'clean' experiment
- when she passes this option in the config file to `True`, this parameter is reflected in the `Scenario` object. This is a parameter that applies either within the scope of a given contributivity measurement method (e.g. pure Shapley), or globally to all contributivity measurement methods of the scenario. Maybe that is something that we should decide first. I'd tend to say within the scope of a given contributivity measurement method (so that when there are several of them, they stay independent and comparable)
- when a contributivity measurement method is called, this new attribute of the `Scenario` is passed along.

Does that make sense? Does that raise specific difficulties on which we should brainstorm together?

multi_partner_learning.py
```diff
@@ -144,7 +150,11 @@ def compute_test_score(self):
         # Initialize variables
         model_to_evaluate, sequentially_trained_model = None, None
         if self.learning_approach in ['seq-pure', 'seq-with-final-agg']:
-            sequentially_trained_model = self.generate_new_model()
+            if os.path.isfile(self.init_model_from):
+                logger.debug("Initiate modem with previous coalition")
```
bowni, line 154: Typo "modem" -> "model"

multi_partner_learning.py
```diff
@@ -215,10 +225,31 @@ def compute_test_score(self):
...
+
+        if not os.path.isdir(model_folder):
+            os.makedirs(model_folder)
+            
+        model_weights = []
+        for layer in model_to_save.layers:
+            layer_weights = []
+            for weight in layer.get_weights():
+                layer_weights.append(weight)
+            model_weights.append(layer_weights)
```
bowni, line 249: Why the layer by layer approach, and not just the `model.get_weights()`?

multi_partner_learning.py
```diff
@@ -487,6 +528,38 @@ def init_with_agg_models(self):
...
+        previous_model.compile(
+            loss=keras.losses.categorical_crossentropy,
+            optimizer="adam",
+            metrics=["accuracy"],
+        )
+        
+        partners_model_list = []
+        for partner in self.partners_list:
+            partners_model_list.append(previous_model)
+        return partners_model_list
```
bowni, line 561: There is a catch here. As it is, you create different pointers to the same model. To have different models for each partner, you should create one for each.

multi_partner_learning.py
```diff
@@ -487,6 +528,38 @@ def init_with_agg_models(self):
...
+        
+        previous_model = self.generate_new_model()
+        previous_model.load_weights(self.init_model_from)
+        previous_model.compile(
+            loss=keras.losses.categorical_crossentropy,
+            optimizer="adam",
+            metrics=["accuracy"],
+        )
+
+        return previous_model
```
bowni, line 544: Hmm, this one is really close to `build_model_from_weights()`. Wouldn't it make sense to have only one of those two? For that, attribute `init_model_from` could be considered as and renamed to `weights_from_a_previous_model`. What do you think @celinejacques ?

multi_partner_learning.py
```diff
@@ -270,10 +301,19 @@ def compute_collaborative_round_fedavg(self):
...
+        logger.debug(os.path.isfile(self.init_model_from))
+        
+
         # Starting model for each partner is the aggregated model from the previous mini-batch iteration
         if is_very_first_minibatch:  # Except for the very first mini-batch where it is a new model
-            logger.debug(f"(fedavg) Very first minibatch of epoch n°{epoch_index}, init new models for each partner")
-            partners_model_list_for_iteration = self.init_with_new_models()
+            if os.path.isfile(self.init_model_from):
+                partners_model_list_for_iteration = self.init_with_previous_learned_models()
+                logger.debug(f"(fedavg) Very first minibatch of epoch n°{epoch_index}, init models with previous coalition model for each partner")
```
bowni, line 313: 2 remarks here:

1. I am not too comfortable with doing the test on the existence of the file. Why not on a `Scenario` attribute corresponding to the option of activating or not this optimization?
1. This should also be applied to `compute_collaborative_round_seqavg()`

scenario.py
bowni: @natct10 I think here we are facing a potential exception: if the parameter is not in `params`. No?

utils.py
```diff
@@ -52,25 +52,48 @@ def get_scenario_params_list(config):
...
                     and (scenario['partners_count'] != len(scenario['samples_split_option'])):
                 raise Exception("Length of samples_split_option does not match number of partners.")
-
+        
             if 'corrupted_datasets' in params_name:
                 if scenario['partners_count'] != len(scenario['corrupted_datasets']):
                     raise Exception("Length of corrupted_datasets does not match number of partners.")
-
+            
+            # keep only init model from if it is a path.
```
bowni, line 88: "If it has a path"? You mean if it is not empty?

utils.py
```diff
@@ -52,25 +52,48 @@ def get_scenario_params_list(config):
...
+        if isinstance(list_scenario['dataset_name'], dict):
+            for dataset_name in list_scenario['dataset_name'].keys():
+                # Add path to init model from an existing model
+                dataset_scenario = list_scenario.copy()
+                dataset_scenario['dataset_name'] = [dataset_name]
+                dataset_scenario['init_model_from'] = list_scenario['dataset_name'][dataset_name]
+                
+                config_dataset.append(dataset_scenario)
+        else:
+            config_dataset.append(list_scenario)
```
bowni, line 68: So the purpose of this is to sort all `list_scenarios` per dataset, before creating the `scenarios` themselves and checking the possible exceptions. I don't see where this sorting per dataset is used, why it is needed?

config.yml
```diff
@@ -2,8 +2,10 @@ experiment_name: my_example
 n_repeats: 5
 scenario_params_list:
  - dataset_name:
-     - 'mnist'
-     - 'cifar10'
+     - 'mnist':
+      -
+     - 'cifar10':
+      -
```
bowni, line 8: What was the reasons for making the `init_from_model` parameter a sub-parameter of the datasets? Why not a global parameter?
Oh, I might see: because you wanted the user to be able to activate it only on some datasets? I think this can make sense, but it would make it simpler if we considered that for doing that, the user just have to list another set of scenarios below with its other dataset. No?

scenario.py
natct10: Yep!

config.yml
celinejacques: I make it a sub-parameter as the models for different datasets are not built with the same architecture. If you give a path for one of the model used on a dataset, it won't work on another one. 
Obviously I can do it as list of different scenario to make it clearer. But in that case, we lose part of the 'factorization' thing to write scenarios easier and quicker. I guess it is worth thinking about.

utils.py
celinejacques: I did that to correctly used the cartesian product and to avoid mixing models between datasets.

config.yml
celinejacques: In fact, finally it could be interesting to have path to previously learned models regarding architecture and not databases. For the moment we only use one model per dataset but in the future, we could use several models for one datasets. I'll include that in this work. What do you think ?

multi_partner_learning.py
celinejacques: I'll have a look on it.

multi_partner_learning.py
```diff
@@ -448,29 +494,48 @@ def build_model_from_weights(self, new_weights):
...
         partners_model_list = []
 
         # Generate a new model and add it to the list
-        new_model = self.generate_new_model()
-        partners_model_list.append(new_model)
+        if use_weights_from_previous_coalition:
+            new_model = self.generate_new_model()
+            new_model.load_weights(self.init_model_from)
+            model_weights = new_model.get_weights()
+            partners_model_list.append(new_model)
```
bowni, line 510: Here it seems that the model is added to the list before being compiled with its new weights. Or does `load_weights()` replaces `set_weights() + .compile()` ?

multi_partner_learning.py
```diff
@@ -448,29 +494,48 @@ def build_model_from_weights(self, new_weights):
...
+            partners_model_list.append(self.build_model_from_weights(model_weights))
 
         return partners_model_list
+    
+    
+    def init_with_model(self, use_weights_from_previous_coalition=False):
+        new_model = self.generate_new_model()
+        
+        if use_weights_from_previous_coalition:
+            new_model.load_weights(self.init_model_from)
```
bowni, line 528: Here also, same question. Why not just use `build_model_from_weights()`?

scenario.py
```diff
@@ -149,6 +149,14 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1, is_dry_ru
         else:
             self.is_early_stopping = True  # default
 
+        # Model from previous coalition used to initialise model
+        self.use_weights_from_previous_coalition = False
+        try:
+            self.init_model_from = params["init_model_from"]
+            self.use_weights_from_previous_coalition = True
+        except:
+            self.init_model_from= ""
```
bowni, line 158: Is there an interest in setting it to an empty string? Couldn't we just leave it undefined? (`pass`...)

multi_partner_learning.py
```diff
@@ -208,10 +220,35 @@ def compute_test_score(self):
...
+            
+        # model_weights = []
+        # for layer in model_to_save.layers:
+        #     layer_weights = []
+        #     for weight in layer.get_weights():
+        #         layer_weights.append(weight)
+        #     model_weights.append(layer_weights)
+            
+        np.save(os.path.join(model_folder, self.dataset_name+'_final_weights.npy'), 
+                model_weights)
```
bowni, line 251: Why both `.h5` and `.npy` formats?

celinejacques: I was thinking that to draw the layer weights it could be easier. But not sure. I can remove the .npy saving.

README.md
```diff
@@ -138,7 +140,8 @@ Example: `n_repeats: 2`
 ##### Choice of dataset
 
 `dataset_name`: `'mnist'` (default) or `'cifar10'`  
-MNIST and CIFAR10 are currently supported. They come with their associated modules in `/datasets` for loading data, pre-processing inputs, and define a model architecture.
+MNIST and CIFAR10 are currently supported. They come with their associated modules in `/datasets` for loading data, pre-processing inputs, and define a model architecture.\
+For each dataset, it is possible to provide a path to model weights learned from a previous coalition. Use `'init'` if you want a random initialization.
```
RomainGoussault, line 144: maybe replace `init `by `random_initialization `?

README.md
```diff
@@ -68,8 +68,10 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
     n_repeats: 5
     scenario_params_list:
      - dataset_name:
-         - 'mnist'
-         - 'cifar10'
+       'mnist':
+        - 'init'
+       'cifar10':
+        - 'init'
```
RomainGoussault, line 74: Is it possible to not put init, and if we do that, by default it will do a random initialization ?

multi_partner_learning.py
```diff
@@ -38,7 +40,10 @@ def __init__(self,
         # Attributes related to the data and the model
         self.val_data = (dataset.x_val, dataset.y_val)
         self.test_data = (dataset.x_test, dataset.y_test)
+        self.dataset_name = dataset.name
         self.generate_new_model = dataset.generate_new_model
+        self.init_model_from = init_model_from
+        self.use_weights_from_previous_coalition = use_weights_from_previous_coalition
```
RomainGoussault, line 46: maybe rename to something like `use_saved_weights`. Or do we assume these weights come from a previous MPL experiment and not from elsewhere in the internet ?

README.md
celinejacques: And by the way, it is still possible to use the older way to enter dataset then the random initialization is applied. I added the possibility to enter dataset as a dictionary key with an empty value.

utils.py
```diff
@@ -51,26 +51,52 @@ def get_scenario_params_list(config):
...
                 if scenario['partners_count'] != len(scenario['corrupted_datasets']):
                     raise Exception("Length of corrupted_datasets does not match number of partners.")
-
+            
+            # keep only init_model_from if a path is given.
+            try:
+                if scenario['init_model_from'] == 'random_initialization':
+                    del scenario['init_model_from']
+            except KeyError:
+                pass
```
bowni, line 96: Ok... question: why not create it in the first place? (L65)

multi_partner_learning.py
```diff
@@ -435,42 +480,57 @@ def aggregate_model_weights(self):
...
     def build_model_from_weights(self, new_weights):
         """Generate a new model initialized with weights passed as arguments"""
 
         new_model = self.generate_new_model()
         new_model.set_weights(new_weights)
-        new_model.compile(
-            loss=keras.losses.categorical_crossentropy,
-            optimizer="adam",
-            metrics=["accuracy"],
-        )
```
bowni, line 448: Are you sure after a `set_weights()` it is ok to not compile? I thought @RomainGoussault had told me you had to, but I might be wrong.

multi_partner_learning.py
```diff
@@ -435,42 +480,57 @@ def aggregate_model_weights(self):
...
 
         return new_model
-
-    def init_with_new_models(self):
+    
+    
+    def init_with_models(self, use_saved_weights=False):
```
bowni, line 494: `use_saved_weights` is an attribute, you could get it without passing it as an argument here, couldn't you?

multi_partner_learning.py
celinejacques: Compile does not impact the weights. You have to compile again if you change the loss or the optimizer for example. We already compiled the model in `generate_new_model()` so we do not have to do it again. 
https://stackoverflow.com/questions/47995324/does-model-compile-initialize-all-the-weights-and-biases-in-keras-tensorflow/47996024

If we want to reinitialize optimizer or loss, we should compile again.

config.yml
```diff
@@ -31,7 +33,8 @@ scenario_params_list:
    dataset_proportion:
      - 1
  - dataset_name:
-     - 'mnist'
+    'mnist':
+    - 'init'
```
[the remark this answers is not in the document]
celinejacques: You're right, I forgot to modify it.

utils.py
celinejacques: The block implemented L65 write the list of parameters so that the Cartesian product can be applied cleanly.

multi_partner_learning.py
celinejacques: Right!

README.md
```diff
@@ -69,9 +69,17 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
...
          - 'mnist'
          - 'cifar10'
        partners_count:
+=======
+       'mnist':
+        - 'random_initialization'
+       'cifar10':
+        - 'random_initialization'
+       partners_count: 
+>>>>>>> 22844f2... Update the Readme file
```
bowni, line 82: A small and simple unresolved conflict @celinejacques :smiley:

scenario.py
```diff
@@ -179,6 +180,14 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1, is_dry_ru
...
             self.is_early_stopping = True  # default
 
+        # Model used to initialise model
+        if params["init_model_from"] == "random_initialization":
+            self.init_model_from = params["init_model_from"]
+            self.use_saved_weights = False
+        else:
+            self.init_model_from = params["init_model_from"]
+            self.use_saved_weights = True
+
```
bowni, line 190: In both cases you have `self.init_model_from = params["init_model_from"]`, so wouldn't it be better to have that before the conditional statement?

multi_partner_learning.py
```diff
@@ -435,36 +468,57 @@ def aggregate_model_weights(self):
...
-        partners_model_list.append(new_model)
+        if self.use_saved_weights:
+            new_model = self.generate_new_model()
+            new_model.load_weights(self.init_model_from)
+            model_weights = new_model.get_weights()
+            partners_model_list.append(new_model)
+        else:
+            new_model = self.generate_new_model()
+            partners_model_list.append(new_model)
+            model_weights = new_model.get_weights()
```
bowni, line 497: You could have 
```
            model_weights = new_model.get_weights()
            partners_model_list.append(new_model)
```
after the conditional statement

bowni (commented):
Seems all good to me, modulo the 3 very minor and quick optimizations I suggest!

multi_partner_learning.py
```diff
@@ -137,7 +142,13 @@ def compute_test_score(self):
         # Initialize variables
         model_to_evaluate, sequentially_trained_model = None, None
         if self.learning_approach in ['seq-pure', 'seq-with-final-agg']:
-            sequentially_trained_model = self.generate_new_model()
+            if self.use_saved_weights:
+                logger.info(f"(seq) Init models with previous coalition model for each partner")
+            else:
+                logger.info(f"(seq) Init new models for each partner")
```
RomainGoussault, line 148: Why do we log that in the `if self.learning_approach in ['seq-pure', 'seq-with-final-agg'] `?

RomainGoussault (commented):
Nice work @celinejacques, super clean !

multi_partner_learning.py
```diff
@@ -208,10 +219,26 @@ def compute_test_score(self):
...
+    def save_final_model_weights(self, model_to_save):
+        """Save final model weights"""
+        
+        model_folder = os.path.join(self.save_folder, 'model')
+
+        if not os.path.isdir(model_folder):
+            os.makedirs(model_folder)
+            
+        model_to_save.save_weights(os.path.join(model_folder, self.dataset_name+'_final_weights.h5'))
+        model_weights = model_to_save.get_weights()
```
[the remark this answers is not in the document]
RomainGoussault: this will be dealt in the logistic regression PR @aygalic

multi_partner_learning.py
```diff
@@ -356,10 +386,13 @@ def compute_collaborative_round_seqavg(self):
 
         # Starting model for each partner is the aggregated model from the previous collaborative round
         if is_very_first_minibatch:  # Except for the very first mini-batch where it is a new model
-            logger.debug(f"(seqavg) Very first minibatch, init a new model for the round")
-            model_for_round = self.generate_new_model()
+            if self.use_saved_weights:
+                logger.info(f"(seqavg) Very first minibatch of epoch n°{epoch_index}, init model with previous coalition model for each partner")
+            else:
+                logger.info(f"(seqavg) Very first minibatch of epoch n°{epoch_index}, init new model for each partner")
+            model_for_round = self.init_with_model()
```
RomainGoussault, line 393: If i understand correctly, this method `compute_collaborative_round_seqavg()` is called at every epoch, which means at every epoch we start with a model from `init_with_model()`. And it looks like `init_with_model()` always returns the same model because the weights `self.init_model_from` are never updated. Am I missing something here ?

multi_partner_learning.py
RomainGoussault: Ok understood

RomainGoussault (commented):
Just one thing I don't understand !

multi_partner_learning.py
bowni: 🤓

bowni (approved):
Looks good to me! (although I confess I didn't do proper tests on my machine)

multi_partner_learning.py
RomainGoussault: It's on the `is_very_first_minibatch`, my bad !

RomainGoussault (approved):
Well done @celinejacques  !!

Issue #168, opened 2020-06-18
Refactor and add section on README on how the validation and test set are handled


RomainGoussault:
- How do we use the validation data set ? For early stopping only ? Is it local or global ?

- Do we have a local or global test set ?

- etc

RomainGoussault:
Looks like val_data is actually a subset of the train set, so it's a local validation_data ? https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/e5ca82f641b32aa7d61199875b09c8fa0a099864/dataset.py#L38

bowni:
I confirm that currently the partner-level validation and test datasets are not used.

RomainGoussault:
Proposition of the use of local/global validation and test sets (I will not discussion implementation details here). 

- **locally** (on each partner node), we don't need both a validation and a test set. We only need one dataset to locally evaluate the model. We won't take any decision (like early stopping) on this evaluation.

- **globally**, we need both a validation and a test set: a validation set to perform early stopping and a test set to evaluate the model.




Other discussion: how the validation/test set are constructed ?

- local validation set: same distribution as the local training set.
- global validation/test set: same distribution as the original dataset. For instance, consider the following simulation on MNIST 2 partners. Parner A has only the digit 3, Partner B has only the digit 6. The test set should contain the 10 digits and not only the digits 3 and 6.

bowni:
So, status to date @RomainGoussault:

1. Locally:
   - the `Partner` object has attributes for both a val and a test set
   - these are populated by `split_data()`, with an SKL `train_test_split()` once data samples are splitted per partners (so with the same distribution as the partner's data)
   - --> to me we are good
1. Globally:
   - the `Scenario` object has a `Dataset` object attribute which in turn has both a val and a test set
   - the test set is populated at data loading, the val set is populated at `Scenario` initialization with an SKL `train_test_split()` (so with the same distribution as the original dataset)
   - --> to me we are good

Do you see something that need to be improved/fixed?

bowni:
I'll add a note in the `README` to describe all that

RomainGoussault:
Thanks @bowni for the recap!

I thinks we are good.

Issue #126, opened 2020-04-28
Amount_per_partner is used twice in the results csv file 


RomainGoussault:
Here: https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/d9727913fd41d7ab1bd4d5e84a00a9f035f30445/scenario.py#L396             
And here: https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/d9727913fd41d7ab1bd4d5e84a00a9f035f30445/scenario.py#L416

The first one is a list of the fractions of dataset used by each partner. --> maybe we should rename it to something like `dataset_fraction_per_partner`.
The second one is a float corresponding to the actual fraction of dataset used by the partner.

bowni:
Why do you think this is a bug? The second one is part of the parameters specific to the partner for which a line is written down. What did you have in mind that we should update/improve?

RomainGoussault:
Ok when I wrote the issue I did NOT notice there was one amount with an S and one without an S..

Anyway, even without my sloppy reading, I think this is can be confusing when analysing the result file. 
But I guess this is not as urgent as I thought it was :)

PR #170, opened 2020-06-18
Housekeeping! (clean samples_split_option, clean local val and test sets, misc. minor improvements)

bowni:
The way `samples_split_option` worked wasn't very clean, as it could take different types of value depending on being initialized to a basic split approach (`'random'` or `'stratified'`) or to an advanced approach (`[[7, 'specific'], [3, 'shared'], [2, 'shared']]`).

Now it looks better:

- it has to be declared as an array with the first value being the type (`'basic'` or `'advanced'`) and the second value the actual split chosen
- the test in `utils.py` for the scenarios combination is done on the type; same for the test in `main.py` for which data splitting function to call
- --> it makes it clean and easy to add new split approaches and even maybe new **types of approaches**

---

Also as suggested in #168, some cleaning with how local val and test sets are populated and used:

- The creation of local val sets is removed from the `preprocess_scenarios_data()` function
- The unused config parameter `single_partner_test_mode` is removed
- Attributes of object `Partner` for local val and test sets are kept
- Creating these val and test sets is now part of the data split functions. For the moment this is basically a `train_test_split()` on each partner's train set

---

Additionnally:

- add a few comments for readibility
- addresses #126

### DIFF
--- a/main.py
+++ b/main.py
@@ -137,7 +137,7 @@ def validate_scenario_list(scenario_params_list, experiment_path):
         current_scenario = scenario.Scenario(scenario_params, experiment_path, is_dry_run=True)
         current_scenario.instantiate_scenario_partners()
 
-        if isinstance(current_scenario.samples_split_option, list):
+        if isinstance(current_scenario.samples_split_description, list):
             current_scenario.split_data_advanced(is_logging_enabled=False)
         else:
             current_scenario.split_data(is_logging_enabled=False)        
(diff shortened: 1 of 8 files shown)


RomainGoussault:
It's strange, it looks like the travis build was not automatically lauched for this PR ?

scenario.py
```diff
@@ -587,8 +591,8 @@ def to_dataframe(self):
         dict_results["train_data_samples_count"] = len(self.dataset.x_train)
         dict_results["test_data_samples_count"] = len(self.dataset.x_test)
         dict_results["partners_count"] = self.partners_count
-        dict_results["amounts_per_partner"] = self.amounts_per_partner
-        dict_results["samples_split_option"] = self.samples_split_option
+        dict_results["dataset_fraction_per_partner"] = self.amounts_per_partner
+        dict_results["samples_split_description"] = self.samples_split_description
```
RomainGoussault, line 595: 👍

main.py
```diff
@@ -137,7 +137,7 @@ def validate_scenario_list(scenario_params_list, experiment_path):
         current_scenario = scenario.Scenario(scenario_params, experiment_path, is_dry_run=True)
         current_scenario.instantiate_scenario_partners()
 
-        if isinstance(current_scenario.samples_split_option, list):
+        if isinstance(current_scenario.samples_split_description, list):
```
RomainGoussault, line 140: `if self.samples_split_type == "advanced"`
(like you did in `run_scenario()`)

bowni: Oh, I missed that one

PR #176, opened 2020-07-02
Switch to tf 2.2

GauthierGar:
change requirements.txt pour utiliser tensorflow2.2.0, ok avec python3.8.2
@natct10

Issue #171, opened 2020-06-22
Update to pandas 1.X


RomainGoussault:
This is needed for google collab --> https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/164#issuecomment-646052648

cc @natct10

natct10:
Okay, thank you Romain! Have you already tested 1.x versions of pandas?

RomainGoussault:
> Okay, thank you Romain! Have you already tested 1.x versions of pandas?

no

PR #178, opened 2020-07-07
Add imdb dataset   a text classification example

GauthierGar:
Need to make some tests with different config.yml file
This branch could be merge with #176 (and maybe #171 also)


config_test_imdb.yml
```diff
@@ -0,0 +1,21 @@
...
+n_repeats: 1
+scenario_params_list:
+ - dataset_name:
+    - 'imdb'
+   partners_count:
+    - 3
+   amounts_per_partner: 
+    - [0.2, 0.5, 0.3]
+   samples_split_option:
+    - ['advanced', [[1, 'shared'], [1, 'shared'], [1, 'specific']]]
```
Thomas-Galtier, line 11: The use of advance split options are not very clear for me, so just to be sure : here you only have two classes, so the two first partners share some sample of class 0 and class 1, and  the last partner has only class 1, am I right?

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,72 @@
...
+        data_list.append(sample)
+
+    data_array = np.array(data_list)
+    data_array = data_array.reshape(data_array.shape[0], data_array.shape[1], 1)
+
+    return data_array
+
+def preprocess_dataset_labels(y): # Do not preprocess labels here! => it's done in another script
+    y = keras.utils.to_categorical(y, num_classes)
+    y = y.reshape(y.shape[0], y.shape[1])
```
Thomas-Galtier, line 37: What is line 37 doing? It seems like it is reshaping to the exact same shape.

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Hidden - Layers
+    model.add(layers.Dropout(0.3, noise_shape=None, seed=None))
+    model.add(layers.Dense(50, activation = "relu"))
+    model.add(layers.Dropout(0.2, noise_shape=None, seed=None))
+    model.add(layers.Dense(50, activation = "relu"))
+    model.add(layers.Flatten()) # Adapt shape of output
+
+    # Output- Layer
+    model.add(layers.Dense(2, activation = "sigmoid")) # 2 => One-Hot Encoding!! with categorical_crossentropy
```
Thomas-Galtier, line 66: Should it rather be a "softmax" activation here ?

config_test_imdb.yml
GauthierGar: Yes, there is only two differents label to caracterize the reviews: positive or negative. But the advanced split options doesn't work like this. The data samples from the initial dataset are split in cluster per data labels.
From the README.txt file: `'advanced' approach [[nb of clusters (int), 'shared' or 'specific']]`
Partners 1 and 2 share a cluster that contains samples of class 0, Partner 3 has the cluster with the samples of class 1.

There is a line in scenario.py to prevent wrong config.yml:
In Scenario.py: ligne 297: `assert specific_clusters_count + shared_clusters_count <= nb_diff_labels`

datasets/dataset_imdb.py
GauthierGar: This line is useless, I removed it

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,72 @@
...
+(x_train, y_train), (x_test, y_test) = imdb.load_data(num_words=nb_words)
+# Argument num_words => only the num_words first more frequent words are used
+# The others are replaced by 2
+# A Review start by the number 1
+# The longest review is less than 3000 words long
+
+def preprocess_dataset_inputs(data):
+    data_list = []
+    for sample in data:
+        sample += [2] * (nb_words - len(sample)) # 2 is the over_of_vocabulary number
```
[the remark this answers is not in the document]
GauthierGar: The words used in the reviews are replaced by integer from 3 to ... , the words 3 ('a') is the most frequent word in the whole dataset, the following is 4 ('of').
1 is the first integer of every review, used in the pre-standardization of the raw text in tf.keras.
2 is used to replaced a word that we don't want to take into account, and I used it to obtain same length vectors.
I don't know if it biasing the model created, because we also have to pre-process the sentence that the model will analyze.
The `load_data()` function can be used to replaced the most and the less frequent words by the `out_of_vocabulary` number. Maybe I should use it?
It seems very complicated: how should I know which word (based on its frequency) is relevant?

config_test_imdb.yml
Thomas-Galtier: Ok its a bit clearer now, thanks

datasets/dataset_imdb.py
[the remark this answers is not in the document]
Thomas-Galtier: ah ok. Sorry I had not read all [the documentation of the `load_data` function](https://keras.io/api/datasets/imdb/#loaddata-function) and, the beginning saying "3" codes for the third most present word, it miss-led me: I thought "1" was coding for the most used word "2" was for the second most used and so on. You can forget my comment. 

Though, yes, it seems interesting to filter the most common words as they do not contains much info. You can use the `get_word_index` function to identify which index is associated to which word (see [the documentation](https://keras.io/api/datasets/imdb/#getwordindex-function)) .

datasets/dataset_imdb.py
GauthierGar: Yes it seems better with categorical_crossentropy, I fixe it

datasets/dataset_imdb.py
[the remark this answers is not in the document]
GauthierGar: I wrote a mistake: [1,2,3,4] code for [the,and,of,a]
The best things to do seems to use 0, as the out_of_vocab, and replaced the most and the less frequent  by 0.
Is it a mistake to use 0? As 0 isn't in the dictionary.

Here's the 100 most common words: `['the', 'and', 'a', 'of', 'to', 'is', 'br', 'in', 'it', 'i', 'this', 'that', 'was', 'as', 'for', 'with', 'movie', 'but', 'film', 'on', 'not', 'you', 'are', 'his', 'have', 'he', 'be', 'one', 'all', 'at', 'by', 'an', 'they', 'who', 'so', 'from', 'like', 'her', 'or', 'just', 'about', "it's", 'out', 'has', 'if', 'some', 'there', 'what', 'good', 'more', 'when', 'very', 'up', 'no', 'time', 'she', 'even', 'my', 'would', 'which', 'only', 'story', 'really', 'see', 'their', 'had', 'can', 'were', 'me', 'well', 'than', 'we', 'much', 'been', 'bad', 'get', 'will', 'do', 'also', 'into', 'people', 'other', 'first', 'great', 'because', 'how', 'him', 'most', "don't", 'made', 'its', 'then', 'way', 'make', 'them', 'too', 'could', 'any', 'movies']`
The best is to stop right before the 36th word: 'like'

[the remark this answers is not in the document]
bowni: > This branch could be merge with #176 (and maybe #171 also)

@GauthierGar :

- Usually we don't consider merging two open PRs. It makes things simpler to consider PRs independently
- #171 is an issue, that is a given topic that we want to work on. A PR addresses an issue (or several), fixes it. But merging a PR with an issue wouldn't mean much :)

bowni:
> This branch could be merge with #176 (and maybe #171 also)

@GauthierGar :

- Usually we don't consider merging two open PRs. It makes things simpler to consider PRs independently
- #171 is an issue, that is a given topic that we want to work on. A PR addresses an issue (or several), fixes it. But merging a PR with an issue wouldn't mean much :)

config_test_imdb.yml
bowni, line 11: Hmmm, the experiment you configured is quite particular:

- You want 2 partners to share samples with 1 label, and a 3rd partner to have samples with another label
- But there are only 2 different labels in this dataset, let's call them `true` and `false` to make it simple here
- This means you have 2 partners with only `true` data samples, and 1 with only `false` data sample
- Learning a model in that configuration is not an obvious thing

I'd strongly suggest that you pick a simpler configuration (a `['basic', 'random'] for example) for validating your implementation.

datasets/dataset_imdb.py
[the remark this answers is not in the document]
bowni: This depends upon the arguments you pass to `load_data()`, like in `imdb.load_data(num_words=10000, skip_top=35, start_char=1, oov_char=2, index_from=3)`.

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Argument num_words => only the num_words first more frequent words are used
+# The others are replaced by 2
+# A Review start by the number 1
+# The longest review is less than 3000 words long
+
+def preprocess_dataset_inputs(data):
+    data_list = []
+    for sample in data:
+        sample += [2] * (nb_words - len(sample)) # 2 is the over_of_vocabulary number
```
bowni, line 27: It would be cleaner if:

- you create a variable for `2` as the oov character, so that you don't have a hardcoded `[2]` and a hardcoded `oov_char=2`
- you define `nb_words` before the definition of this function (for readibility of the code). Maybe rename it to something easier to understand (like `max_nb_words_configured`? something very explicit)

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,71 @@
...
+
+def preprocess_dataset_labels(y): # Do not preprocess labels here! => it's done in another script
+    y = keras.utils.to_categorical(y, num_classes)
+
+    return y
+
+# Padding
+nb_words = 3000
+
+input_shape = (nb_words,1)
```
bowni, line 43: Hmmm, maybe add a check in the padding function that verifies that the longest review is shorter than the `nb_words` parameter you impose.

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Hidden - Layers
+    model.add(layers.Dropout(0.3, noise_shape=None, seed=None))
+    model.add(layers.Dense(50, activation = "relu"))
+    model.add(layers.Dropout(0.2, noise_shape=None, seed=None))
+    model.add(layers.Dense(50, activation = "relu"))
+    model.add(layers.Flatten()) # Adapt shape of output
+
+    # Output- Layer
+    model.add(layers.Dense(2, activation = "softmax")) # 2 => One-Hot Encoding!! with categorical_crossentropy
```
bowni, line 65: Does this model architecture make sense? I am not experienced enough with NN to get an opinion on the spot, but I'd suggest verifying/challenging it.

datasets/dataset_imdb.py
GauthierGar: This is done

datasets/dataset_imdb.py
GauthierGar: There is now an `assert` in the function `preprocess_dataset_inputs`

datasets/dataset_imdb.py
GauthierGar: The model didn't make any sens, There is a new one with quite good performance, I am currently training it :)

datasets/dataset_imdb.py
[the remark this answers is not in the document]
GauthierGar: I have completely change this function
The new pre-processing of inputs is quite arbitrary but give good results.
We only keep the 10 000 more frequent words (the others are replaced by the out_of_vocabulary integer = 2), and the 200 first words of the reviews.
If the `len(review) < 200` we add 0 at the start to obtain a review of 200 words.

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,60 @@
...
+https://keras.io/examples/nlp/text_classification_from_scratch/
+https://builtin.com/data-science/how-build-neural-network-keras
+"""
+
+import tensorflow as tf
+import numpy as np
+from tensorflow import keras
+from keras.utils import to_categorical
+from keras import models
+from keras import layers
```
aygalic, line 14: You may want to use a single line for this : 
from keras import models, layers

GauthierGar: Done, the imports seem clean:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/a0d52796ba6cf1519947a4b2b70fa5e022ec1dd0/datasets/dataset_imdb.py#L9-L13

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,60 @@
...
+import keras
+from keras.utils import to_categorical
+from keras.preprocessing.sequence import pad_sequences
+from keras import models, layers
+from keras.datasets import imdb
+
+out_of_vocabulary = 2
+# We keep the 10000 first more frequent words in the datasets
+# Also the dimentionality of the input of the Embedding layer
+size_vocabulary = 20000
```
bowni, line 17: Two comments here:

1. Unclear if this is 10000 or 20000
1. Just for readibility, maybe something more like:
   ```py
   # Definition of dataset-specific parameters
   size_vocabulary = ...  # Explanation...
   maxlen = ...  # Explanation...
   ...
   ```

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,60 @@
...
+num_classes = 2
+
+# Not used here, but important another module
+input_shape = keras.Input(shape=(None,), dtype="int32")
+
+(x_train, y_train), (x_test, y_test) = imdb.load_data(num_words=size_vocabulary, skip_top=35, start_char=1, oov_char=out_of_vocabulary, index_from=3)
+
+# We skip the 35 most frequent words
+# The others are replaced by 2
+# A Review start by the number 1
```
bowni, line 30: Nice that you documented that. Maybe put that next to the above dataset-specific parameter definition, it doesn't make too much sense to have them apart

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,60 @@
...
+# We keep the 10000 first more frequent words in the datasets
+# Also the dimentionality of the input of the Embedding layer
+size_vocabulary = 20000
+# We only keep the 200 first words of the reviews
+maxlen = 200
+# Number of labels
+num_classes = 2
+
+# Not used here, but important another module
+input_shape = keras.Input(shape=(None,), dtype="int32")
```
bowni, line 24: Hmmm, where and how is this used actually?

config_test_imdb.yml
```diff
@@ -0,0 +1,21 @@
+experiment_name: test_imdb
```
natct10, line 1: Is it possible to remove this file and to add the dataset to `config.yaml` and `config_quick_debug.yml` file, in `scenario_params_list` / `dataset_name` please?

datasets/dataset_imdb.py
GauthierGar: I don't know, but an error is raised if input_shape isn't defined, like `preprocess_dataset_labels()`.
I will search where this come from

config_test_imdb.yml
GauthierGar: Done, I had to create a new scenario: there is only 2 labels in the imdb dataset, so we can't create more than 2 clusters

config.yml
```diff
@@ -49,3 +49,23 @@ scenario_params_list:
...
+   aggregation_weighting:
+    - 'uniform'
+   methods:
+    - ["Shapley values", "Independent scores"]
+   epoch_count:
+    - 10
+   minibatch_count:
+    - 8
+   gradient_updates_per_pass_count:
+    - 5
```
bowni, line 71: Adding `'imdb'` after line 32 would be a better factorization wouldn't it? 😃

config_quick_debug.yml
```diff
@@ -20,3 +20,21 @@ scenario_params_list:
...
+   multi_partner_learning_approach:
+    - 'fedavg'
+   aggregation_weighting:
+    - 'uniform'
+   methods:
+    - ["Shapley values", "Independent scores"]
+   gradient_updates_per_pass_count:
+    - 5
+   is_quick_demo:
+    - True
```
bowni, line 40: I wouldn't put it in the debug config. It has to stay a debug config. I think we should also remove CIFAR10 from there

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,52 @@
...
+
+import keras
+from keras.utils import to_categorical
+from keras.preprocessing.sequence import pad_sequences
+from keras import models, layers
+from keras.datasets import imdb
+
+# Definition of dataset-specific parameters 
+maxlen = 200 # We only keep the 200 first words of the reviews
+out_of_vocabulary = 2
```
bowni, line 16: Let's provide an explicative comment about that one (otherwise no one can understand what this is :)!

Also, I think there is PEP8 thing for inline comments: two whitespaces before the `#`

GauthierGar: `Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.` yes, it's done

config.yml
GauthierGar: That seems to easy, It's probably a trap :rotating_light:

config_quick_debug.yml
```diff
@@ -3,7 +3,6 @@ n_repeats: 1
 scenario_params_list:
  - dataset_name:
     - 'mnist'
```
RomainGoussault, line 5: I would leave at least 2 datasets for this config, so at least 2 datasets will be run by Travis.

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,52 @@
...
+def preprocess_dataset_inputs(x):
+    return pad_sequences(x_train, maxlen=maxlen)
+
+x_train = preprocess_dataset_inputs(x_train)
+x_test = preprocess_dataset_inputs(x_test)
+
+def generate_new_model_for_dataset():
+    model = models.Sequential()
+
+    model = models.Sequential()
```
RomainGoussault, line 43: to remove (same line as above)

datasets/dataset_imdb.py
```diff
@@ -0,0 +1,52 @@
...
+    model.add(layers.Embedding(size_vocabulary, 128))  # Embed each integer in a 128-dimensional vector
+    model.add(layers.Bidirectional(layers.LSTM(64, return_sequences=True)))
+    model.add(layers.Bidirectional(layers.LSTM(64)))
+
+    model.add(layers.Dense(2, activation="softmax"))
+
+    model.compile(optimizer = "adam", loss = "categorical_crossentropy", metrics = ["accuracy"])
```
RomainGoussault, line 50: no whitespace for arguments: `optimizer="adam", loss="categorical_crossentropy",` ...
https://www.python.org/dev/peps/pep-0008/#other-recommendations

config_quick_debug.yml
GauthierGar: Done, cifar10 is back for travis

GauthierGar:
> What is the performance on 2 or3 partners ?
My computer isn't able to run it properly, @bowni trained the model and stopped it around 70% (before the end of the computation)

> Do you have the same issue on your machine ?
Yes, I have the same warning message but the computation starts. I think it could come from 2 things:
This one:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/b3710e33a10b290b750c23b53caaf3494665d75f/datasets/dataset_imdb.py#L43
 I found some docs:
https://stackoverflow.com/questions/35892412/tensorflow-dense-gradient-explanation
https://github.com/tflearn/tflearn/issues/460
Or from this one:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/b3710e33a10b290b750c23b53caaf3494665d75f/datasets/dataset_imdb.py#L25

I need to what is the purpose of `input_shape`

bowni:
I had this too but it actually computed. I tried on 30% of the data over 2 partners (0,1 and 0,2), 20 mini-batches, 10 gradient updates per pass. I passed 70% acc on val data at approx the 10th mini-batch of the 1st epoch, and stopped there.

bowni:
I confirm @RomainGoussault that with the below config I passed 0.7 at 9th minibatch of 1st epoch.
Note that I relaxed the assert requiring the `amounts_per_partner` to total 1, and that in fact it didn't really work because of the way `split_data()` is implemented. We need rebase `x_train` to the total amount selected (here `0.3`) before splitting it. But this is for another PR, the one with the asserts @GauthierGar.

```yaml
experiment_name: debug_config
n_repeats: 1
scenario_params_list:
 - dataset_name:
    - 'imdb'
   partners_count:
    - 2
   amounts_per_partner: 
    - [0.2, 0.1]
   samples_split_option:
    - ['basic', 'random']
    - 'fedavg'
   aggregation_weighting:
    - 'uniform'
    - 5
   minibatch_count:
    - 20
```

datasets/dataset_imdb.py
GauthierGar: `input_shape` is an argument of the constructor in dataset.py but It seems to stay unused during the computation.
Should we remove this feature? I don't know, maybe it be can be usefull in some cases

GauthierGar:
I have removed the 2nd compilation of the model with categorical-CE and the adam optimizer to solve #181

RomainGoussault:
It now runs better on my machine. I am currently running on experient on 3 partners, looks OK.


@GauthierGar you can use collab to run it if it is too slow on your machine --> https://colab.research.google.com/github/SubstraFoundation/distributed-learning-contributivity/blob/master/run_experiment_on_google_collab.ipynb

bowni:
@RomainGoussault so we let you conclude your tests, and if OK then merge?

RomainGoussault:
> @RomainGoussault so we let you conclude your tests, and if OK then merge?

Yes !

RomainGoussault:
I have strange results, test score is 0.5 for all 4 methods of mpl.


[imdb_pr_2020-07-20_11h39.zip](https://github.com/SubstraFoundation/distributed-learning-contributivity/files/4952396/imdb_pr_2020-07-20_11h39.zip)

bowni:
Now moved to PR #238 and very close to be concluded and merged!

PR #180, opened 2020-07-08
Add the PVRL method

Thomas-Galtier:
PVRL stands for partner value by reinforcment learning.


RomainGoussault:
How different is it from the DVRL approach ?

Thomas-Galtier:
The DVRL approach assigns a value to each data, whereas the PVRL approach assigns a value to the dataset of each partner.

Thomas-Galtier:
also there is no need to compute gradients with tensorflow in the PVRL approach which makes it much easier to implement

contributivity.py
```diff
@@ -962,6 +963,83 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
+                    save_folder="",
+                    )
+        mpl.compute_test_score()
+        weights_for_starting_model = mpl.federated_model_weights
+        previous_loss = mpl.loss_collective_models[-1]
+        t=0
+        while (t<the_scenario.epoch_count*the_scenario.partners_count or np.sum(np.abs(partner_values -previous_partner_values))/ the_scenario.partners_count > epsilon ):
+            t+=1
+            print("t:",t)
+            print("partner_values:",partner_values)
```
bowni, line 993: Better to use the loggers with debug level instead of `print` statements 😃

multi_partner_learning.py
```diff
@@ -56,10 +57,11 @@ def __init__(self,
         self.minibatched_y_train = [None] * self.partners_count
         self.aggregation_weights = []
         self.models_weights_list = [None] * self.partners_count
+        self.federated_model_weights = weights_for_starting_model
         self.scores_last_learning_round = [None] * self.partners_count
         self.score_matrix_per_partner = np.nan * np.zeros(shape=(self.epoch_count, self.minibatch_count, self.partners_count))
         self.score_matrix_collective_models = np.nan * np.zeros(shape=(self.epoch_count, self.minibatch_count + 1))
-        self.loss_collective_models = []
+        self.loss_collective_models = []
```
bowni, line 64: OK so for `multi_partner_learning.py` there seem to be a lot to adapt. I'd say:

- Adding a `loss_collective_models` table: seems needed, good addition
- Adding support for starting with an existing model: this has been done more in depth by @celinejacques in a recently merged PR. It would be best to remove this implementation, and adapt to what @celinejacques has done

Does that seem OK @Thomas-Galtier?

bowni (changes requested):
Hi @Thomas-Galtier, this looks very promising and I must say I am quite excited about benchmarking it vs. Shapley (and others) in different setups!
But first I think it requires adaptation to what we more recently merged into `master`, in particular @celinejacques implementation of starting with an existing model.

multi_partner_learning.py
Thomas-Galtier: Yep that seems ok

bowni:
Followed-up in PR #241

PR #182, opened 2020-07-09
Add output messages with the asserts and tests of the config.yml: unkown parameters

### LINKED ISSUE #163: Assert should output useful messages
Most of the assertion in the scenario constructor do not provide messages. We should provide some messages to add some context of why these assertions are done

GauthierGar:
In order to resolve #163

### DIFF
--- a/scenario.py
+++ b/scenario.py
@@ -29,7 +29,7 @@ class Scenario:
 
         # Raise Exception if unknown parameters in the .yml file
 
-        params_known = ["dataset_name","partners_count","amounts_per_partner","samples_split_option","multi_partner_learning_approach","aggregation_weighting","methods","gradient_updates_per_pass_count","is_quick_demo"]
+        params_known = ["dataset_name","partners_count","amounts_per_partner","samples_split_option","multi_partner_learning_approach","aggregation_weighting","methods","gradient_updates_per_pass_count","is_quick_demo", "corrupted_datasets", "epoch_count", "minibatch_count", "is_early_stopping"]
 
         if not all([x in params_known for x in params]):
             for x in params:
@@ -423,7 +423,6 @@ class Scenario:
         # Then we parameterize this via the splitting_indices to be passed to np.split
         # This is to transform the percentages from the scenario configuration into indices where to split the data
         if self.partners_count == 1:
-            splitting_indices = self.amounts_per_partner[0]
             splitting_indices_train = 1
         else:
             splitting_indices = np.empty((self.partners_count - 1,))
--- a/scenario.py
+++ b/scenario.py
@@ -454,7 +454,8 @@ class Scenario:
             splitting_indices = splitting_indices / dataset_fraction_used  # Normalize proportions
 
             if self.partners_count == 1:
-                splitting_indices_train = 1  # np.split(train_idx, splitting_indices_train=1) returns train_idx
+                splitting_indices_train = 1
+                # np.split(train_idx, splitting_indices_train=1) returns train_idx
             else :
                 # splitting_indices_train (argument of np.split) can be an integer or a np.array
                 splitting_indices_train = (splitting_indices * len(x_train)).astype(int)
--- a/scenario.py
+++ b/scenario.py
@@ -138,19 +138,19 @@ class Scenario:
         # Number of epochs, mini-batches and fit_batches in ML training
         if "epoch_count" in params:
             self.epoch_count = params["epoch_count"]
-            assert self.epoch_count > 0, "Error: in .yml epoch_count should be > 0"
+            assert self.epoch_count > 0, "Error: in the provided config file, epoch_count should be > 0"
         else:
             self.epoch_count = 40  # default
 
         if "minibatch_count" in params:
             self.minibatch_count = params["minibatch_count"]
-            assert self.minibatch_count > 0, "Error: in .yml minibatch_count should be > 0"
+            assert self.minibatch_count > 0, "Error: in the provided config file, minibatch_count should be > 0"
         else:
             self.minibatch_count = 20  # default
 
         if "gradient_updates_per_pass_count" in params:
             self.gradient_updates_per_pass_count = params["gradient_updates_per_pass_count"]
-            assert self.gradient_updates_per_pass_count > 0, "Error: in .yml gradient_updates_per_pass_count should be > 0"
+            assert self.gradient_updates_per_pass_count > 0, "Error: in the provided config file, gradient_updates_per_pass_count should be > 0"
         else:
             self.gradient_updates_per_pass_count = constants.DEFAULT_GRADIENT_UPDATES_PER_PASS_COUNT
 
@@ -393,7 +393,7 @@ class Scenario:
             )
 
         # Check coherence of number of mini-batches versus partner with small dataset
-        assert self.minibatch_count <= min([len(p.x_train) for p in self.partners_list]), "Error: in .yml and the dataset provided, a partner doesn't have enough data samples to create the minibatches "
+        assert self.minibatch_count <= min([len(p.x_train) for p in self.partners_list]), "Error: in the provided config file and the provided dataset, a partner doesn't have enough data samples to create the minibatches "
 
         if is_logging_enabled:
             logger.info("### Splitting data among partners:")
@@ -419,8 +419,8 @@ class Scenario:
         # ... or receive different amounts?
 
         # Check the percentages of samples per partner and control its coherence
-        assert len(self.amounts_per_partner) == self.partners_count, "Error: in .yml in amounts_per_partner arguments: the amounts_per_partners arguments (a list) has to contain a proportion per each partner - Here there is too many or not enough proportions"
-        assert np.sum(self.amounts_per_partner) <= 1, "Error: in .yml in amounts_per_partner arguments: the sum of the proportions you provided is higher than 1"
+        assert len(self.amounts_per_partner) == self.partners_count, "Error: in the provided config file, amounts_per_partner list should have a size equals to partners_count"
+        assert np.sum(self.amounts_per_partner) <= 1, "Error: in the provided config file, in amounts_per_partner arguments: the sum of the proportions you provided is higher than 1"
 
         # Then we parameterize this via the splitting_indices to be passed to np.split
         # This is to transform the percentages from the scenario configuration into indices where to split the data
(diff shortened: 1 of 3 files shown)


GauthierGar:
There are 2 strange tests in `split_data` and `split_data_advanced` (called in main.py):

https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/93212d1970965fe670999c6438bd1ab5140419db/scenario.py#L487-L488
This seems to be a mistake: `self.minibatch_count` is the number (the user chooses it in config.yml) of minibatch of data: 10? 20?
and `(min(self.amounts_per_partner) * len(x_train))` is the size of the smallest dataset (from the partner with the smallest proportion of data over the whole dataset)
I think that we want to make sure that the size of a minibatch `len(x_train) / self.minibatch_count` is lower than the size of the smallest dataset
So I suggested writing instead: `assert self.minibatch_count <= (min(self.amounts_per_partner) * len(x_train))`

https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/93212d1970965fe670999c6438bd1ab5140419db/scenario.py#L385-L386
This is exactly the same, so I suggest the same change.

Moreover, shouldn't we try to factorize all these tests in one new function?

RomainGoussault:
I think this test is still valid: we want to check if the number of minibatch is smaller than the number of data samples of one partner. For instance, we don't want to be in the the situation where we have `minibatch_count = 100` and one partner that has only 50 data samples.

bowni:
**Topic 1**

> I think this test is still valid: we want to check if the number of minibatch is smaller than the number of data samples of one partner. For instance, we don't want to be in the the situation where we have `minibatch_count = 100` and one partner that has only 50 data samples.

I confirm that it was my intent when I wrote this assert :smiley: 

---

**Topic 2**

> So I suggested writing instead: `assert self.minibatch_count <= (min(self.amounts_per_partner) * len(x_train))`
> 
> https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/93212d1970965fe670999c6438bd1ab5140419db/scenario.py#L385-L386
> 
> This is exactly the same, so I suggest the same change.

In your suggestion @GauthierGar you suppose that the `Partner` object has already been instantiated for each partner, and that the original dataset has already been split among them. So the test your propose could work, but it should be placed downstream, at the end of the `split_data...` functions (`split_data()` and `split_data_advanced()`).
It would be an interesting one, not exactly the same actually than the original test. When data are split in the 'advanced' way, in many cases not all samples from the original dataset are distributed among partners. So we might encounter cases where the first test passes, but this second one don't.

---

**Topic 3**

> Moreover, shouldn't we try to factorize all these tests in one new function?

Do you mean all asserts? That is a good question, I don't what is the best practice here / the most Pythonic way to handle that: all asserts in one place, or asserts along the code as close as possible to the particular topic they handle? @natct10 @RomainGoussault

scenario.py
```diff
@@ -292,7 +294,7 @@ def split_data_advanced(self, is_logging_enabled=True):
             shared_clusters_count = max([p.cluster_count for p in partners_with_shared_clusters])
         else:
             shared_clusters_count = 0
-        assert specific_clusters_count + shared_clusters_count <= nb_diff_labels
+        assert specific_clusters_count + shared_clusters_count <= nb_diff_labels, "Error: data samples from the initial dataset are split in clusters per data labels - ['advanced', [[7, 'shared'], [6, 'shared'], [2, 'specific'], [1, 'specific']]] means 7 shared clusters and 2 + 1 = 3 specific clusters ==> The dataset need to have >= 10 labels"
```
bowni, line 297: You can't write "10 labels" as it depends upon the dataset configured, try to word it in a more generic way.
Also, good thing to give an example, just indicate explicitely that you are providing an example :smiley:

scenario.py
```diff
@@ -407,8 +409,8 @@ def split_data(self, is_logging_enabled=True):
         # ... or receive different amounts?
 
         # Check the percentages of samples per partner and control its coherence
-        assert len(self.amounts_per_partner) == self.partners_count
-        assert np.sum(self.amounts_per_partner) == 1
+        assert len(self.amounts_per_partner) == self.partners_count, "Error: in .yml in amounts_per_partner arguments: the amounts_per_partners arguments (a list) has to contain a proportion per each partner - Here there is too many or not enough proportion"
+        assert np.sum(self.amounts_per_partner) == 1, "Error: in .yml in amounts_per_partner arguments: the proportions you provided don't sum up to exactly 1"
```
bowni, line 413: I just had an idea: we should relax this test to `<=1` instead of `==1`. If a user wants to use only 40% of the dataset, that is fine, and maybe sometimes even a good idea.

RomainGoussault:
> Moreover, shouldn't we try to factorize all these tests in one new function?

We can do that, but we should keep in mind that it's is easier to read if the assert statements are close to variables they are trying to check.

natct10:
> > Moreover, shouldn't we try to factorize all these tests in one new function?
> 
> We can do that, but we should keep in mind that it's is easier to read if the assert statements are close to variables they are trying to check.

I agree with you @RomainGoussault.
Thank you @GauthierGar for the error messages with the assertions, this is way cleaner!

GauthierGar:
I will use this branch to add tests focusing on the config.yml to solve #184

natct10:
> I will use this branch to add tests focusing on the config.yml
> to solve #184

Great! You might be interested in this PR then: https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/160

GauthierGar:
> Great! You might be interested in this PR then: #160 
Is it something that should be solved using unit testing?
The aim is to raise an error before the computation if there is an unrecognized parameter in the .yml
I thought that unit testing was about making automatic tests of the effectiveness of each function, not about checking if the parameters are valid.

bowni:
> > Great! You might be interested in this PR then: #160
> 
> Is it something that should be solved using unit testing?
> The aim is to raise an error before the computation if there is an unrecognized parameter in the .yml
> I thought that unit testing was about making automatic tests of the effectiveness of each function, not about checking if the parameters are valid. @natct10

I think you are right @GauthierGar, but I let @natct10 confirm that it is better practice to have proper unit tests on one side, and controls like this one on the other side.

natct10:
> > > Great! You might be interested in this PR then: #160
> > 
> > 
> > Is it something that should be solved using unit testing?
> > The aim is to raise an error before the computation if there is an unrecognized parameter in the .yml
> > I thought that unit testing was about making automatic tests of the effectiveness of each function, not about checking if the parameters are valid. @natct10
> 
> I think you are right @GauthierGar, but I let @natct10 confirm that it is better practice to have proper unit tests on one side, and controls like this one on the other side.

You are right, it would be better to test the validity of the config file here and raise errors as you did @GauthierGar, which is different from unit testing.

GauthierGar:
I have implemented a new feature: the user can use a part < 100% of the dataset in the basic split option. (your last comment about the PR #178 @bowni )
We keep random element of `x_train`, this could be problematic with the `stratified` mode if one label is push out of the data.

This could be implemented too in the advanced split option, but it will be more complex

bowni:
> I have implemented a new feature: the user can use a part < 100% of the dataset in the basic split option. (your last comment about the PR #178 @bowni )
> We keep random element of `x_train`, this could be problematic with the `stratified` mode if one label is push out of the data.
> 
> This could be implemented too in the advanced split option, but it will be more complex

Excellent initiative! I'll look into it later today or tomorrow @GauthierGar

scenario.py
```diff
@@ -126,19 +138,19 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1, is_dry_ru
...
             self.epoch_count = params["epoch_count"]
-            assert self.epoch_count > 0
+            assert self.epoch_count > 0, "Error: in .yml epoch_count should be > 0"
         else:
             self.epoch_count = 40  # default
 
         if "minibatch_count" in params:
             self.minibatch_count = params["minibatch_count"]
-            assert self.minibatch_count > 0
+            assert self.minibatch_count > 0, "Error: in .yml minibatch_count should be > 0"
```
RomainGoussault, line 147: I would replace` Error: in .yml `by `Error in the provided config file, minibatch ...`
This is the same for the other assert messages

scenario.py
```diff
@@ -407,12 +419,13 @@ def split_data(self, is_logging_enabled=True):
         # ... or receive different amounts?
 
         # Check the percentages of samples per partner and control its coherence
-        assert len(self.amounts_per_partner) == self.partners_count
-        assert np.sum(self.amounts_per_partner) == 1
+        assert len(self.amounts_per_partner) == self.partners_count, "Error: in .yml in amounts_per_partner arguments: the amounts_per_partners arguments (a list) has to contain a proportion per each partner - Here there is too many or not enough proportions"
```
RomainGoussault, line 422: Would replace by `"amounts_per_partner list should have a size equals to partners_count"`

scenario.py
```diff
@@ -423,14 +436,32 @@ def split_data(self, is_logging_enabled=True):
                 )
             splitting_indices_train = (splitting_indices * len(y_train)).astype(int)
 
+        # If partners don't use the full dataset we only keep the amount of data targeted and then normalize proportions
+
+        proportion_of_dataset = np.sum(self.amounts_per_partner)
```
RomainGoussault, line 441: Would rename to smth like: dataset_fraction_used

scenario.py
```diff
@@ -423,14 +436,32 @@ def split_data(self, is_logging_enabled=True):
...
+
+        if proportion_of_dataset < 1:
+            skip_idx = int(round(len(x_train) * proportion_of_dataset))
```
RomainGoussault, line 444: Add a log.info() for this case ?

scenario.py
```diff
@@ -423,14 +436,32 @@ def split_data(self, is_logging_enabled=True):
...
+            train_idx = np.arange(len(x_train))
+            np.random.seed(42)
+            np.random.shuffle(train_idx)
+            x_train = x_train[train_idx[0:skip_idx]]
+            y_train = y_train[train_idx[0:skip_idx]]
+
+            splitting_indices = splitting_indices / proportion_of_dataset  # Normalize proportions
+            if self.partners_count == 1:
+                splitting_indices_train = 1  # Read Documentation of np.split
```
RomainGoussault, line 453: Could you elaborate :) ?

GauthierGar: I removed this line because it's useless,
I wrote the comment because I was surprised `np.split` could take as a argument, an array or an integer(1).
`np.split(train_idx, 1)` return the array train_idx

RomainGoussault: Could you write this explanation directly in the comment?

GauthierGar:
I think that this PR is ready to be merged,
I have removed a feature (the one that keep a little part of the dataset, <100%), I will add it in another PR as discussed this morning.

Issue #174, opened 2020-06-25
Reduce the number of files created for experiments


RomainGoussault:
We already did some works in how the results files are handled https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/96.

However I think we should go one step further and remove the number of files created. For instance in the experiment --> https://github.com/SubstraFoundation/distributed-learning-contributivity/tree/portfolio/saved_experiments/mnist_cifar10_distributed_learning there around 800 files created.

PR #183, opened 2020-07-10
Investigate castastrophic forgetting

RomainGoussault:


constants.py
```diff
@@ -7,7 +7,7 @@
 DEFAULT_BATCH_SIZE = 256
 MAX_BATCH_SIZE = 2**20
 DEFAULT_GRADIENT_UPDATES_PER_PASS_COUNT = 8
-PATIENCE = 4  # patience for early stopping
+PATIENCE = 30  # patience for early stopping
```
bowni, line 10: :open_mouth:

RomainGoussault:
image
Hard to know why sometimes it works sometimes it does not. Nothing in the logs.
I am running more experients with less parameters variation and more repeats.

RomainGoussault:
The notebook should be put in the saved_experiments folder. Also the data should be pushed to the repo.

arthurPignet:
> The notebook should be put in the saved_experiments folder. Also the data should be pushed to the repo.

What do you mean by the data should be pushed to the repo ? Not the dataset ?

arthurPignet (commented):
Really interesting 
The notebook works with the main file, in the old functionnement of the library, don't you think it could be interesting to use the new mplc package ? 
with a list of the scenario tested, etc
However it is not mandatory for merging

arthurPignet (approved):

arthurPignet (changes requested):
On the notebook : 
Maybe you could add few sentences on the choice of your parameters. Why one category per partner, why these tests on the minibatch parameter.

Why do you print the 40 graphics ? And only for the scenario with one minibatch ?

What you are naming local_validation_score is computed on the global validation set, with the local model, you can have a confusion here. (Which is important, because technically with catastrophic forgetting, the local model overfit on the local train set. Each partner got only one class, so the accuracy on the the local validation dataset would be near 100%)

bowni:
@RomainGoussault are @arthurPignet comments relevant from your perspective? Can you indicate if you plan to keep making updates to this PR or if we should merge it now?

RomainGoussault:
> > The notebook should be put in the saved_experiments folder. Also the data should be pushed to the repo.
> 
> What do you mean by the data should be pushed to the repo ? Not the dataset ?

I meant, the results files should be pushed to the repos. Actually I did pushed the results.csv but I forget to push the individual results. (When we resolved this https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/174, life will be easier :))

RomainGoussault:
@arthurPignet you are right on the local_validation_score, it is global_actually. (we should rename the array `score_matrix_per_partner`, it's a bit confusing).
And as you said, it will be nice to have the local validation score, but I think we don't have it.

RomainGoussault:
@bowni yes I still plan to improve the notebooks, sorry for the delay !.
And Thanks @arthurPignet for the very relevant comments !

The next steps are:
* pushing all the results (all the folders to get the loss data)
* see if we can get the local validation score but I think we can not.

bowni (approved):

arthurPignet (approved):

PR #185, opened 2020-07-13
Adding Titanic Dataset

aygalic:
# Initial commit of new branch
## Adding support for a new dataset.

The Titanic dataset is based on tabular data which differ from the other dataset used in this project.

Inspired by : 
https://web.stanford.edu/class/archive/cs/cs109/cs109.1166/problem12.html


scenario.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1, is_dry_ru
         # ---------------------------------------------------------------------
 
         # Get and verify which dataset is configured
-        supported_datasets_names = ["mnist", "cifar10"]
+        supported_datasets_names = ["mnist", "cifar10","titanic"]
```
bowni, line 31: Missing a whitespace after the comma 😃

datasets/dataset_titanic.py
```diff
@@ -0,0 +1,84 @@
...
+
+    raw_dataset = pd.read_csv('./train.csv', index_col=False )
+    x = raw_dataset.drop(['Name','Survived'], axis=1)
+    x = preprocess_dataset_inputs(x)
+    y = raw_dataset['Survived']
+    y = y.to_numpy()
+
+    x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=0.33, random_state=42)
+
+    os.remove("train.csv") # not the cleanest way to deal with temp file
```
bowni, line 58: 2 points here:

1. If this isn't the cleanest way, what would it be and can we implement it? 😃 
1. 2 whitespaces before inline comments says PEP8!

datasets/dataset_titanic.py
```diff
@@ -0,0 +1,84 @@
...
+    model.add(Dense(128,activation="relu"))
+    model.add(keras.layers.core.Dropout(0.2))
+    model.add(Dense(64,activation="relu"))
+    model.add(Dense(2,activation="softmax"))
+    model.compile(
+        loss="categorical_crossentropy",
+        optimizer="adam",
+        metrics=["accuracy"]
+    )
+    return model
```
bowni, line 81: Where does this architecture comes from? Always good to indicate the source, if it is a reference one.

aygalic: I added the kaggle notebook from which I got the architecture, I hope it makes sense.

datasets/dataset_titanic.py
aygalic: Thanks for the reference, i will take care of not committing the same mistake.

I also changed the import of the dataset. It's now directly done through read_csv().

RomainGoussault:
Nice work @aygalic !

Could you tell us if we have a decent performance on a scenario with 3 partners, random split ?

You can run it in collab if it takes too long on your marchine --> https://colab.research.google.com/github/SubstraFoundation/distributed-learning-contributivity/blob/master/run_experiment_on_google_collab.ipynb

aygalic:
> You can run it in collab if it takes too long on your marchine --> https://colab.research.google.com/github/SubstraFoundation/distributed-learning-contributivity/blob/master/run_experiment_on_google_collab.ipynb



> Nice work @aygalic !
> 
> Could you tell us if we have a decent performance on a scenario with 3 partners, random split ?
> 
> You can run it in collab if it takes too long on your marchine --> https://colab.research.google.com/github/SubstraFoundation/distributed-learning-contributivity/blob/master/run_experiment_on_google_collab.ipynb

Thank you ! 

After my last commit, I did a GPU run on colab with the following config :  

experiment_name: my_example
n_repeats: 1
scenario_params_list:
 - dataset_name:
     - 'titanic'
   partners_count:
     - 3
   amounts_per_partner:
     - [0.4, 0.3, 0.3]
   samples_split_option:
     - ['basic', 'random']
     - 'seq-pure'
   aggregation_weighting:
     - 'data_volume'
     - 'uniform'
   epoch_count:
     - 20
   methods:
     - ["Shapley values", "Independent scores", "TMCS"]
   minibatch_count:
     - 10
     - 8

It took  around 22 min to run the main.py file

config.yml
```diff
@@ -1,51 +1,23 @@
 experiment_name: my_example
```
bowni, line 1: I think we talked about it already: you have to keep `config.yml` as is (5 repeats, MNIST and CIFAR10, etc.). If you want, you can add Titanic to it.

multi_partner_learning.py
```diff
@@ -441,12 +441,6 @@ def build_model_from_weights(self, new_weights):
 
         new_model = self.generate_new_model()
         new_model.set_weights(new_weights)
-        new_model.compile(
-            loss=keras.losses.categorical_crossentropy,
-            optimizer="adam",
-            metrics=["accuracy"],
-        )
```
bowni, line 448: Are you sure that after setting new weights it is not needed to compile? I am not sure about it, this would require a confirmation from Keras' documentation, or someone else who is sure.

bowni:
> 2. I tried a few combinations quickly but couldn't find easily what would make sense to reach a good performance (in terms of `minibatch_count`, type of multi-partner learning and aggregation...). I always plateau at `0.6`, which I suppose is the dummy performance. Could you manage to reach a relatively good perf and add as comments what is the right config for it?

@aygalic could you update the `config_quick_debug.yml` with parameters than enable a good perf with the model you implemented for this dataset? This will enable us to test by ourselves before merging.

config.yml
aygalic: Yes, I will revert the `config.yml` file.
Since the performance are not great with this model I rather keep it as an available exemple rather than a technical demo.

bowni:
@aygalic you'll also have to update the `README` to indicate that Titanic joined the family of supported datasets!

multi_partner_learning.py
aygalic: As long as the model is compiled once in the dedicated dataset file, I guess we should be fine.
I try to show it with this experiment.
[Model recompiling.zip](https://github.com/SubstraFoundation/distributed-learning-contributivity/files/5002498/Model.recompiling.zip)

This post on stack overflow seems to go in the same dirrection
https://stackoverflow.com/questions/47995324/does-model-compile-initialize-all-the-weights-and-biases-in-keras-tensorflow/47996024

According to keras's documentation : 
https://keras.io/api/models/model_training_apis/#compile-method

The `compile()` method seems to only "Configures the model for training." which doesn't help a lot.

aygalic:
> > 1. I tried a few combinations quickly but couldn't find easily what would make sense to reach a good performance (in terms of `minibatch_count`, type of multi-partner learning and aggregation...). I always plateau at `0.6`, which I suppose is the dummy performance. Could you manage to reach a relatively good perf and add as comments what is the right config for it?
> 
> @aygalic could you update the `config_quick_debug.yml` with parameters than enable a good perf with the model you implemented for this dataset? This will enable us to test by ourselves before merging.

Yes ! Both done.

I added my configuration at the end of the quick debug file. 
A run worked for me

multi_partner_learning.py
aygalic: [ab_testing_recompile.zip](https://github.com/SubstraFoundation/distributed-learning-contributivity/files/5006267/ab_testing_recompile.zip)

From this testing, it seems to be fine aswell

this is the config used : 

```
experiment_name: ab_testing
n_repeats: 1
scenario_params_list:
 - dataset_name:
     - 'mnist'
   partners_count:
     - 3
   amounts_per_partner:
     - [0.4, 0.3, 0.3]
   samples_split_option:
     - ['basic', 'random']
     - 'fedavg'
   aggregation_weighting:
     - 'uniform'
   epoch_count:
     - 10
   methods:
     - ["Shapley values"]
   minibatch_count:
     - 10
     - 8

```

I've also set random seeds in the main file : 

```
from numpy.random import seed
seed(1)
tf.compat.v1.set_random_seed(
    1
)
```

bowni: Ok great. The StackOverflow answer clarifies it clearly. Good news then, we can remove a good lot of unnecessary `.compile()`.

PR #188, opened 2020-07-16
Adding scikit learn random forest support

aygalic:
Code "working" : an entire round is possible without error with the correct config file. 
The config file is named : config_test.yml.
It introduces a new class of "multi_partner_learning_approach" which is : 'fed-forest'.

The goal here is to use models such as random forest in order to go through an entire collaborative learning process.

The learning process doesn't seems to work properly. We need further investigation with this and the federated learning part.


RomainGoussault:
How random forest are supposed to work in a federated setting ?
You iteratively add a tree on each partner ?

aygalic:
> How random forest are supposed to work in a federated setting ?
> You iteratively add a tree on each partner ?

Iteratively adding tree on each partner might be a good approach, I have no clue on whether or not it will provide any enhancement over non-federated learning.

The following article seems to use partial trees for federated learning.
https://arxiv.org/abs/1905.10053
Unfortunately, it works with a centralised configuration which isn't compatible with the substra framework.

I was also thinking about a vote system for each partner, but I don't really know how to implement this and whether or not it can provide decent improvement over non federated learning. I believe some kind of ensemble learning-based approach might do the trick.

Also, I want to investigate further with different approaches of ensemble learning provided by sklearn :
https://scikit-learn.org/stable/modules/ensemble.html

I'm not 100% any of these approaches leads to the right solution, but I hope a better support for scikit learn in general can be beneficial.
Maybe working with bagging classifier might be a better option.

Issue #187, opened 2020-07-15
Unused argument and functions in dataset.py


GauthierGar:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/3414c86ec432abf79ead3334ec101a74a1999f9a/dataset.py#L38-L48

These 2 functions are never used/called in the whole repo, we maybe should removed them

https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/3414c86ec432abf79ead3334ec101a74a1999f9a/dataset.py#L25
this is why the user has to create a variable `input_shape` in datasets/dataset_xx.py 
this attribute seems to stay unused, we could removed it

GauthierGar:
I wrote a mistake, the function `train_val_split` is used in `scenario.py`, `preprocess_labels` is stil unused.
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/3414c86ec432abf79ead3334ec101a74a1999f9a/dataset.py#L38-L42

In `scenario.py`, the training data are often split (in 2 times) with `train_test_split(self.x_train, self.y_train, test_size=0.2, random_state=42)` to obtain test and validation data:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/3414c86ec432abf79ead3334ec101a74a1999f9a/scenario.py#L471-L476

We could improve the readibility by replacing all these lines with a call to `dataset.train_val_split`

bowni:
This is now obsolete - closing it

PR #190, opened 2020-07-17
Unit test in partner.py - cifar10

GauthierGar:
Add unit tests for the 2 modules `partner.py` and `datasets/dataset_cifar10.py`

This could be used as an example to create new tests when adding a feature to the project
I have first wrote test in the `tests.py` and then implement these constraints in the methods `shuffle_labels` and `corrupt_labels` with ` if ... raise Exception`
These `Exception` are then caught by `with pytest.raises(Exception):`

If we plan to add lots of tests, we could consider creating 1 test module for each module to keep it easy to read.

I need to make these tests more automatic. To test all the functions of the projects, we could create a script that will generate a list of scenario, from the datasets and federated learning approaches currenty supported. This list would be used as arguments in `tests.py` to test more complex functions such as `split_data` (Functions that need `scenario` `mpl` `dataset` objects as arguments)

### DIFF
--- a/dataset.py
+++ b/dataset.py
@@ -36,7 +36,7 @@ class Dataset:
         self.generate_new_model_for_dataset = generate_new_model_for_dataset
 
     def train_val_split(self):
-        # this could be add in the constructor??
+        """Called once, after Dataset's constructor"""
         if self.x_val != None or self.y_val != None:
             raise Exception
 
@@ -45,7 +45,7 @@ class Dataset:
         )
 
     def preprocess_labels(self):
-        # should be removed
+
         self.y_train = self.preprocess_dataset_labels(self.y_train)
         self.y_val = self.preprocess_dataset_labels(self.y_val)
         self.y_test = self.preprocess_dataset_labels(self.y_test)
--- a/dataset.py
+++ b/dataset.py
@@ -36,13 +36,16 @@ class Dataset:
         self.generate_new_model_for_dataset = generate_new_model_for_dataset
 
     def train_val_split(self):
+        # this could be add in the constructor??
+        if self.x_val != None or self.y_val != None:
+            raise Exception
 
         self.x_train, self.x_val, self.y_train, self.y_val = train_test_split(
             self.x_train, self.y_train, test_size=0.2, random_state=42
         )
 
     def preprocess_labels(self):
-
+        # should be removed
         self.y_train = self.preprocess_dataset_labels(self.y_train)
         self.y_val = self.preprocess_dataset_labels(self.y_val)
         self.y_test = self.preprocess_dataset_labels(self.y_test)
--- a/dataset.py
+++ b/dataset.py
@@ -44,11 +44,6 @@ class Dataset:
             self.x_train, self.y_train, test_size=0.2, random_state=42
         )
 
-    def preprocess_labels(self):
-
-        self.y_train = self.preprocess_dataset_labels(self.y_train)
-        self.y_val = self.preprocess_dataset_labels(self.y_val)
-        self.y_test = self.preprocess_dataset_labels(self.y_test)
 
     def generate_new_model(self):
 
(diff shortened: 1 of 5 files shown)


GauthierGar:
solve #191

tests.py
```diff
@@ -7,6 +8,207 @@
...
+from datasets import dataset_cifar10 as data_cf
+from datasets import dataset_mnist as data_mn
+from partner import Partner
+from dataset import Dataset
+from scenario import Scenario
+from multi_partner_learning import MultiPartnerLearning
+from pathlib import Path
+
+
+@pytest.fixture(scope="module", params=["a","b"])
```
natct10, line 25: Need space in the params!

natct10: And have you seen the `paramatrize`: 
```python
@pytest.mark.parametrize("a,b,expected", testdata, ids=["forward", "backward"])
def test_timedistance_v1(a, b, expected):
    diff = a - b
    assert diff == expected
```
It might be interesting if you want to mimic some configs: https://docs.pytest.org/en/latest/example/parametrize.html#paramexamples

tests.py
```diff
@@ -7,6 +8,207 @@
...
+    scenar.dataset
+
+    yield scenar
+
+def test_scenar(create_scenario):
+    assert type(create_scenario) == Scenario
+
+@pytest.fixture(scope="class")
+def create_cifar10_x_train():
+    (x_train, y_train), (x_test, y_test) = cifar10.load_data()
```
natct10, line 165: Do you need `(x_test, y_test)` here, as it seems to be unused?

natct10: or maybe test the `train` and `test`?

GauthierGar: To create `x_train` I have to load these 3 variables, I could also make test on `x_test`

I started writing the class `Test_dataset_cifar10` to learn `pytest` - I could maybe keep this class as an example
I will try to implement test on the labels `y_train` and `y_test`.

natct10: Ok great!

tests.py
```diff
@@ -1,12 +1,356 @@
-# Pytest doc: https://docs.pytest.org/en/latest/getting-started.html#create-your-first-test
+# -*- coding: utf-8 -*-
+"""
+This enables to parameterize unit tests - the tests are run by Travis each you commit to the github repo
```
natct10, line 3: each *time

scenario.py
```diff
@@ -269,6 +269,10 @@ def append_contributivity(self, contributivity):
         self.contributivity_list.append(contributivity)
 
     def instantiate_scenario_partners(self):
+        """Create the partners_list - should be []"""
+
+        if self.partners_list != []:
+            raise Exception
```
natct10, line 275: Would be nice to use a message, please!

dataset.py
```diff
@@ -36,6 +36,9 @@ def __init__(self,
         self.generate_new_model_for_dataset = generate_new_model_for_dataset
 
     def train_val_split(self):
+        """Called once, after Dataset's constructor"""
+        if self.x_val != None or self.y_val != None:
```
natct10, line 40: What about `if self.x_val or self.y_val`?

partner.py
```diff
@@ -33,15 +33,28 @@ def __init__(self, partner_id):
         self.y_val = None
         self.y_test = None
 
-    def get_x_train_len(self):
+    def get_x_train_len(self):  # not used
```
natct10, line 36: Indeed, could you please remove it?

dataset.py
```diff
@@ -37,8 +37,8 @@ def __init__(self,
 
     def train_val_split(self):
         """Called once, after Dataset's constructor"""
-        if self.x_val != None or self.y_val != None:
-            raise Exception
+        if self.x_val or self.y_val:
+            raise Exception("x_val and y_val should be of NoneType")
```
natct10, line 41: +1

GauthierGar:
solve #187 
input_shape is still unused

PR #198, opened 2020-07-23
Logistic regression support for any approach

aygalic:
Adding support for logistic regression when using either `seq-pure` or `fedavg `approches.


config.yml
```diff
@@ -1,51 +1,24 @@
 experiment_name: my_example
-n_repeats: 5
+n_repeats: 1
 scenario_params_list:
  - dataset_name:
-     - 'mnist'
-     - 'cifar10'
+     - 'titanic'
    partners_count:
```
bowni, line 6: Hmmm, I see that you needed an updated `config.yml` for your tests, but please keep it close to its current state, maybe just adding the Titanic dataset, as it is the "main" config example of the library.
(If you play with it, just don't commit it!)

config.yml
```diff
@@ -1,51 +1,24 @@
...
+   epoch_count:
+     - 10
    methods:
      - ["Shapley values", "Independent scores", "TMCS"]
-   minibatch_count: 
-     - 20
+   minibatch_count:
+     - 5
    gradient_updates_per_pass_count:
      - 8
```
bowni, line 24: I think that `8` is a lot with that few samples in the dataset...

datasets/dataset_titanic.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Dropping the useless features
+    x.drop('Name', axis=1, inplace=True)
+    x.drop('Pclass', axis=1, inplace=True)
+    x.drop('Siblings/Spouses Aboard', axis=1, inplace=True)
+    x.drop('Parents/Children Aboard', axis=1, inplace=True)
+    x.drop('Title', axis=1, inplace=True)
+    return x.to_numpy()
+
+# download train and test sets
```
bowni, line 51: Small detail: capitalize the first letter!

scenario.py
```diff
@@ -197,17 +199,18 @@ def __init__(self, params, experiment_path, scenario_id=1, n_repeat=1, is_dry_ru
...
+                index_val = np.random.choice(self.dataset.x_val.shape[0], 500, replace=False)
+                index_test = np.random.choice(self.dataset.x_test.shape[0], 500, replace=False)
+                self.dataset.x_train = self.dataset.x_train[index_train]
+                self.dataset.y_train = self.dataset.y_train[index_train]
+                self.dataset.x_val = self.dataset.x_val[index_val]
+                self.dataset.y_val = self.dataset.y_val[index_val]
+                self.dataset.x_test = self.dataset.x_test[index_test]
+                self.dataset.y_test = self.dataset.y_test[index_test]
             self.epoch_count = 3
             self.minibatch_count = 2
```
bowni, line 215: :+1:

multi_partner_learning.py
```diff
@@ -85,23 +87,31 @@ def compute_test_score_for_single_partner(self, partner):
...
+        else:
+            history = model.fit(
+                partner.x_train,
+                partner.y_train,
+                batch_size=partner.batch_size,
+                epochs=self.epoch_count,
+                verbose=0,
+                validation_data=self.val_data,
+                callbacks=cb,
+            )
```
bowni, line 101: Ok, just one comment for the whole `multi_partner_learning.py`:

- Wow, that's a lot, well done @aygalic!
- What is really particularly interesting I think it to work with an SKL model instead of a Keras model. It raises questions: can we make the library compatible/agnostic with different ML backends?
- The `if... else...` everywhere is not ideal. Could you find some factorization for that, a bit like with the example of `collaborative_round_fit()`?

config.yml
aygalic: I had in mind to reverse it to its original state before merging with `master`, maybe just adding the `titanic `dataset. 

I wanted to provide a template of this file for everyone to take a test run on this. It might not be the smartest way to do this. 
Should I just reverse it now and add it to my `.gitignore` ?

config.yml
aygalic: I've changed it to 3, it seems to provide a small improvement in accuracy, but a nice speedup in computing time, which is nice.

aygalic:
> @aygalic can you share the results you get when training on Titanic with this model?

Here are my results

### For seq-pure :
all_partners
federated_training_acc

### For fedavg : 
all_partners
federated_training_acc

config_test.yml
```diff
@@ -0,0 +1,28 @@
+experiment_name: my_example
```
RomainGoussault, line 1: We probably don't need this config file in master.

multi_partner_learning.py
```diff
@@ -480,17 +524,33 @@ def collaborative_round_fit(model_to_fit, train_data, val_data, batch_size):
...
+                batch_size=batch_size,
+                epochs=1,
+                verbose=0,
+                validation_data=val_data,
+            )
         return history
 
+    @staticmethod
+    def collaborative_round_evaluation(model_to_evaluate, evaluation_data):
+        """Evaluate the model with arguments passed as parameters and returns the history object"""
```
RomainGoussault, line 542: Does it return the history for logistic regression ?

multi_partner_learning.py
```diff
@@ -85,21 +87,21 @@ def compute_test_score_for_single_partner(self, partner):
...
+                batch_size=partner.batch_size,
+                epochs=self.epoch_count,
+                verbose=0,
+                validation_data=self.val_data,
+                callbacks=cb,
+            )
 
         # Evaluate trained model
-        model_evaluation = model.evaluate(x_test, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        model_evaluation = self.collaborative_round_evaluation(model, self.test_data)
```
RomainGoussault, line 104: We should precise it's evaluated on the test set, so maybe rename to something like `test_evaluation `or `test_score`

multi_partner_learning.py
```diff
@@ -171,9 +170,9 @@ def compute_test_score(self):
                 model_to_evaluate = sequentially_trained_model
             elif self.learning_approach in ['fedavg', 'seq-with-final-agg', 'seqavg']:
                 model_to_evaluate = self.build_model_from_weights(self.aggregate_model_weights())
-            model_evaluation = model_to_evaluate.evaluate(
-                x_val, y_val, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0,
-            )
+
+
+            model_evaluation = self.collaborative_round_evaluation(model_to_evaluate, self.val_data)
```
RomainGoussault, line 175: Same thing here : `val_evaluation `or `val_score`.
And the same applies in the all the other occurences of calls to `collaborative_round_evaluation()`

multi_partner_learning.py
```diff
@@ -274,7 +273,12 @@ def compute_collaborative_round_fedavg(self):
 
         # Evaluate and store accuracy of mini-batch start model
         model_to_evaluate = partners_model_list_for_iteration[0]
-        model_evaluation = model_to_evaluate.evaluate(x_val, y_val, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+
+        if not hasattr(model_to_evaluate, 'coef_'):
+            model_evaluation = [0]*2
```
RomainGoussault, line 278: I don't understand why we need this bit

multi_partner_learning.py
```diff
@@ -425,22 +435,48 @@ def prepare_aggregation_weights(self):
     def aggregate_model_weights(self):
         """Aggregate model weights from the list of models, with a weighted average"""
 
-        weights_per_layer = list(zip(*self.models_weights_list))
-        new_weights = list()
+        if isinstance(self.generate_new_model(), type(LogisticRegression())):
```
RomainGoussault, line 438: Do we have to generate a new model here ? Can't we use one of the existing models : self.models_weights_list[0] ?

multi_partner_learning.py
```diff
@@ -480,17 +524,33 @@ def collaborative_round_fit(model_to_fit, train_data, val_data, batch_size):
...
-            x_train,
-            y_train,
-            batch_size=batch_size,
-            epochs=1,
-            verbose=0,
-            validation_data=val_data,
-        )
-
+        if isinstance(model_to_fit, type(LogisticRegression())):
+            history = model_to_fit.fit(x_train, y_train)
```
RomainGoussault, line 528: This does not return an history object but return the model. But maybe we should create one from scratch to mimic the output of keras ?

multi_partner_learning.py
RomainGoussault: Ah yes it mimics the history object, that is good !
We should just change the comments to say it returns` [loss, accuracy]`

RomainGoussault: As we said, rename function to `evalute_model`

multi_partner_learning.py
```diff
@@ -364,7 +370,10 @@ def compute_collaborative_round_seqavg(self):
             model_for_round = self.init_with_agg_model()
 
         # Evaluate and store accuracy of mini-batch start model
-        model_evaluation = model_for_round.evaluate(x_val, y_val, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        if not hasattr(model_for_round, 'coef_'):
+            model_evaluation = [0]*2
```
RomainGoussault, line 374: move logic to collaborative_round_evaluation function()

multi_partner_learning.py
```diff
@@ -504,7 +564,12 @@ def log_collaborative_round_partner_result(self, partner, partner_index, validat
     def update_iterative_results(self, partner_index, fit_history):
         """Update the results arrays with results from the collaboration round"""
 
-        validation_score = fit_history.history["val_accuracy"][0]
+        if isinstance(fit_history, type(LogisticRegression())):
+            validation_score = self.collaborative_round_evaluation(fit_history, self.val_data)[0]
+        else :
+            validation_score = fit_history.history["val_accuracy"][0]
+
+
```
RomainGoussault, line 572: remove blank lines

multi_partner_learning.py
```diff
@@ -504,7 +564,12 @@ def log_collaborative_round_partner_result(self, partner, partner_index, validat
     def update_iterative_results(self, partner_index, fit_history):
         """Update the results arrays with results from the collaboration round"""
 
-        validation_score = fit_history.history["val_accuracy"][0]
+        if isinstance(fit_history, type(LogisticRegression())):
+            validation_score = self.collaborative_round_evaluation(fit_history, self.val_data)[0]
```
RomainGoussault, line 568: use history object

multi_partner_learning.py
aygalic: I ended up using `test_evaluation `and `val_evaluation `to avoid conflicts with the `test_score `attribute in `multi_partner_learning.py` file.

multi_partner_learning.py
aygalic: It's needed for non fitted model evalutation which keras does support, but sklearn doesn't.

I move it to `evaluate_model`

multi_partner_learning.py
aygalic: I changed it for 
`if type(self.models_weights_list[0]) is tuple: `
Which is a bit better as keras weights are list

multi_partner_learning.py
aygalic: I did it with a dict with a single value for 
`loss`
`accuracy`
`val_loss`
`val_accuracy`
Hope it improves compatibility

multi_partner_learning.py
```diff
@@ -476,21 +509,49 @@ def init_with_agg_models(self):
...
+    @staticmethod
+    def evaluate_model(model_to_evaluate, evaluation_data):
+        """Evaluate the model with arguments passed as parameters and returns [loss, accuracy]"""
+
+        x_eval, y_eval = evaluation_data
+
+        if isinstance(model_to_evaluate, type(LogisticRegression())):
+            if not hasattr(model_to_evaluate, 'coef_'):
+                model_evaluation = [0]*2
+            else :
```
natct10, line 547: Could you please remove  the space here?

multi_partner_learning.py
```diff
@@ -476,21 +509,49 @@ def init_with_agg_models(self):
...
+                loss = log_loss(y_eval, model_to_evaluate.predict(x_eval))  # mimic keras model evaluation
+                accuracy = model_to_evaluate.score(x_eval, y_eval)
+                model_evaluation = [loss, accuracy]
+        else :
```
natct10, line 551: same here, please

aygalic: done, thank you

datasets/dataset_titanic.py
```diff
@@ -74,12 +78,15 @@ def generate_new_model_for_dataset():
     model.add(Dense(16, activation='relu'))
     model.add(Dense(8, activation='relu'))
     model.add(Dense(1, activation='sigmoid'))
-
     model.compile(optimizer = 'adam',
                    loss = 'binary_crossentropy',
                    metrics = ['accuracy'])
+    """
```
RomainGoussault, line 84: I would remove the part where we build the neural network model since we don't use it

multi_partner_learning.py
```diff
@@ -198,11 +197,12 @@ def compute_test_score(self):
 
         # After last epoch or if early stopping was triggered, evaluate model on the global testset
         logger.info("### Evaluating model on test data:")
-        model_evaluation = model_to_evaluate.evaluate(
-            x_test, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+
+        test_evaluation = self.evaluate_model(model_to_evaluate, self.test_data)
```
RomainGoussault, line 201: 👍  for using test/val names !!

multi_partner_learning.py
```diff
@@ -425,22 +426,46 @@ def prepare_aggregation_weights(self):
     def aggregate_model_weights(self):
         """Aggregate model weights from the list of models, with a weighted average"""
 
-        weights_per_layer = list(zip(*self.models_weights_list))
-        new_weights = list()
+        if type(self.models_weights_list[0]) is tuple:  # Check weights type for aggregation
```
RomainGoussault, line 429: this if/else corresponds to Logistic Regression VS neural networks ? If so, we should at least indicate it in comments

multi_partner_learning.py
```diff
@@ -506,12 +567,12 @@ def update_iterative_results(self, partner_index, fit_history):
 
         validation_score = fit_history.history["val_accuracy"][0]
 
+
```
RomainGoussault, line 570: remove blank line

RomainGoussault (changes requested):
👍 !

multi_partner_learning.py
aygalic: True, I'll detail it above

RomainGoussault (approved):
Super work @aygalic !

datasets/dataset_titanic.py
```diff
@@ -60,26 +62,13 @@ def load_data():
...
-    model.add(Dense(1, activation='sigmoid'))
-
-    model.compile(optimizer = 'adam',
-                   loss = 'binary_crossentropy',
-                   metrics = ['accuracy'])
-
-    return model
+    """Return a LogisticRegression Classifier"""
+
+
```
bowni, line 67: Unnecessary blank line here

datasets/dataset_titanic.py
```diff
@@ -60,26 +62,13 @@ def load_data():
...
+    clf = LogisticRegression(max_iter=10000, warm_start=1, random_state = 0)
```
bowni, line 68: Unecessary whitespaces in `random_state = 0`

multi_partner_learning.py
```diff
@@ -85,26 +87,26 @@ def compute_test_score_for_single_partner(self, partner):
...
+        else:
+            history = model.fit(
+                partner.x_train,
+                partner.y_train,
+                batch_size=partner.batch_size,
+                epochs=self.epoch_count,
+                verbose=0,
+                validation_data=self.val_data,
+                callbacks=cb,
+            )
```
bowni, line 101: It would be best to have this conditional statement in a dedicated fit function, to keep `compute_test_score_for_single_partner()` independent from the model type.

multi_partner_learning.py
```diff
@@ -85,26 +87,26 @@ def compute_test_score_for_single_partner(self, partner):
...
 
         # Evaluate trained model
-        model_evaluation = model.evaluate(x_test, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        test_evaluation = self.evaluate_model(model, self.test_data)
         logger.info(f"   Model evaluation on test data: "
-                    f"{list(zip(model.metrics_names, ['%.3f' % elem for elem in model_evaluation]))}")
+                    f"{list(zip(model.metrics_names, ['%.3f' % elem for elem in test_evaluation]))}")
```
bowni, line 106: I find `model_evaluation` is a clearer name than `test_evaluation`

multi_partner_learning.py
```diff
@@ -171,18 +170,18 @@ def compute_test_score(self):
...
+
+            current_val_loss = val_evaluation[0]
+            current_val_metric = val_evaluation[1]
 
             self.score_matrix_collective_models[epoch_index, minibatch_count] = current_val_metric
             self.loss_collective_models.append(current_val_loss)
 
             logger.info(f"   Model evaluation at the end of the epoch: "
-                        f"{['%.3f' % elem for elem in model_evaluation]}")
+                        f"{['%.3f' % elem for elem in val_evaluation]}")
```
bowni, line 184: Same here, `model_evaluation` is clearer. But I get the iead, you wanted to identify clearly evaluations on val data vs. test data. So let's use explicit names then: `model_eval_val_data` and `model_eval_test_data`.

multi_partner_learning.py
```diff
@@ -425,22 +422,45 @@ def prepare_aggregation_weights(self):
     def aggregate_model_weights(self):
         """Aggregate model weights from the list of models, with a weighted average"""
 
-        weights_per_layer = list(zip(*self.models_weights_list))
-        new_weights = list()
+        # Sklearn models weigths are tuples while keras model's weights are list, this differentiate keras vs sk-learn models
+        if type(self.models_weights_list[0]) is tuple:  # Check weights type for aggregation
```
bowni, line 426: Ok I suppose this works, but it is difficult to understand and maintain. Wouldn't it be clearer/cleaner if you test on the model type? (LogisticRegression)

multi_partner_learning.py
```diff
@@ -451,12 +471,20 @@ def init_with_new_models(self):
...
+            # For each remaining partner, create a new model and add it to the list
+            for i in range(len(self.partners_list)-1):
+                partners_model_list.append(self.generate_new_model())
+        else:
+            partners_model_list.append(new_model)
+
+            # For each remaining partner, duplicate the new model and add it to the list
+            new_model_weights = new_model.get_weights()
+            for i in range(len(self.partners_list)-1):
+                partners_model_list.append(self.build_model_from_weights(new_model_weights))  # is it really necessary to build it from weight instead of scratch ?
```
bowni, line 487: Yes it is necessary: this allows having the exact same weights for each partner, and not have different models to start from (as `generate_new_model()` uses random initialization)

multi_partner_learning.py
```diff
@@ -476,21 +504,49 @@ def init_with_agg_models(self):
...
-            batch_size=batch_size,
-            epochs=1,
-            verbose=0,
-            validation_data=val_data,
-        )
-
+        if isinstance(model_to_fit, type(LogisticRegression())):
+            history = model_to_fit.fit(x_train, y_train)
+            [loss, acc] = MultiPartnerLearning.evaluate_model(model_to_fit, train_data)
+            [val_loss, val_acc] = MultiPartnerLearning.evaluate_model(model_to_fit, val_data)
```
bowni, line 514: Why use this syntax with the name of the class here?

bowni (changes requested):
Looks good, in the right tracks and close to the end!

multi_partner_learning.py
aygalic: Very nice suggestion. It simplifies the logic a lot, the `fit_model()` function was almost already able to handle single partner scenarii.

multi_partner_learning.py
aygalic: I will use this notation

multi_partner_learning.py
aygalic: For this specific instance, the model isn't initialized before (neither after), wich would be expensive for this single test.

I understand this isn't clear but I hope the comment above demystify the situation.

What do you think @bowni ?

multi_partner_learning.py
aygalic: Alright, comment removed ! Thanks

multi_partner_learning.py
aygalic: Because of the `@staticmethod` tag.

Is there any way to do better ?

bowni:
@aygalic there seem to be a last thing for you to do :smiley: It looks like when you adapted to the last merge (Celine's PR) you left one action not adapted to SKL's LogisticRegression: `model_to_save.save_weights(os.path.join(model_folder, self.dataset_name+'_final_weights.h5'))` gives `AttributeError: 'LogisticRegression' object has no attribute 'save_weights'
`
Could you fix that one and then we merge and you are good for some rest? :muscle:

multi_partner_learning.py
bowni: Ok then. We could question the benefit of making it a staticmethod in the first place, but let's not do that we have enough on our plate :smiley:

multi_partner_learning.py
bowni: Ok, I get it. Let's keep it that way for the moment, but it will need some improvement at some point. I opened an issue on that to not forget it.

multi_partner_learning.py
```diff
@@ -209,35 +198,43 @@ def compute_test_score(self):
...
-            
-        np.save(os.path.join(model_folder, self.dataset_name+'_final_weights.npy'), 
+
+        if isinstance(model_to_save, type(LogisticRegression())):
+            dump(model_to_save, os.path.join(model_folder, self.dataset_name+'_final_weights.h5'))
+            coefs = np.array(model_to_save.coef_ )
+            intercepts = np.array(model_to_save.intercept_ )
+
+            np.savez(os.path.join(model_folder, self.dataset_name+'_final_weights.npy'),
+                name1=coefs, name2=intercepts)
```
bowni, line 231: Hmmm, the `savez` seems like a French verb? :laughing:

multi_partner_learning.py
```diff
@@ -458,27 +454,51 @@ def prepare_aggregation_weights(self):
...
+            new_model.n_iter_ = 0
+        else:
+            new_model.set_weights(new_weights)
+        return new_model
 
         return new_model
-    
-    
+
+
```
bowni, line 501: This looks strange: twice `return new_model` and 2 blank lines

multi_partner_learning.py
```diff
@@ -491,24 +511,30 @@ def init_with_models(self):
             new_model.load_weights(self.init_model_from)
```
bowni, line 511: Will `load_weights(self.init_model_from)` work with LogisticRegression @aygalic ?

multi_partner_learning.py
```diff
@@ -491,24 +511,30 @@ def init_with_models(self):
...
+            # For each remaining partner, duplicate the new model and add it to the list
+            new_model_weights = new_model.get_weights()
+            for i in range(len(self.partners_list)-1):
+                partners_model_list.append(self.build_model_from_weights(new_model_weights))
 
         return partners_model_list
-    
-    
+
+
```
bowni, line 531: Just 1 blank line here

multi_partner_learning.py
```diff
@@ -491,24 +511,30 @@ def init_with_models(self):
...
     def init_with_model(self):
         new_model = self.generate_new_model()
-        
+
         if self.use_saved_weights:
             new_model.load_weights(self.init_model_from)
```
bowni, line 536: Same question here: will `load_weights(self.init_model_from)` work with LogisticRegression?

multi_partner_learning.py
bowni: When testing I got `TypeError: _save_dispatcher() got an unexpected keyword argument 'name1'`

bowni (changes requested):
Hey @aygalic , I spotted a few weird things, which I changed on my machine locally to do tests. Unfortunately it breaks, there is a use of arguments in an `np.save()` that seems to be odd. Would you be able to keep fighting this thing?

multi_partner_learning.py
bowni: My mistake of course 🤦‍♂️

multi_partner_learning.py
bowni: Done

multi_partner_learning.py
bowni: Done

bowni (approved):
Works well for me now, well done @aygalic. @RomainGoussault @natct10 could you please review and merge?

config_quick_debug.yml
```diff
@@ -28,4 +28,26 @@ scenario_params_list:
...
+   epoch_count:
+    - 2
+   methods:
+    - ["Shapley values", "Independent scores"]
+   minibatch_count:
+    - 2
+   gradient_updates_per_pass_count:
+    - 2
+   is_quick_demo:
+    - True
```
RomainGoussault, line 53: We don't want to overcharge the quick debug. So consider removing this part. 
However you can add an end-to-end like this one --> https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/master/tests/end_to_end_tests.py

multi_partner_learning.py
```diff
@@ -90,26 +93,15 @@ def compute_test_score_for_single_partner(self, partner):
...
-            callbacks=cb,
-        )
-
-        # Reference the testset according to the scenario configuration
-        x_test, y_test = self.test_data
+        history = self.fit_model(model, (partner.x_train, partner.y_train), self.val_data, partner.batch_size, self.epoch_count)
 
         # Evaluate trained model
-        model_evaluation = model.evaluate(x_test, y_test, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+        model_evaluation_test_data= self.evaluate_model(model, self.test_data)
```
RomainGoussault, line 99: missing space before =

multi_partner_learning.py
```diff
@@ -458,27 +447,47 @@ def prepare_aggregation_weights(self):
...
 
     def build_model_from_weights(self, new_weights):
         """Generate a new model initialized with weights passed as arguments"""
 
         new_model = self.generate_new_model()
-        new_model.set_weights(new_weights)
 
+        if isinstance(new_model, type(LogisticRegression())):
+            new_model.coef_, new_model.intercept_ = new_weights
+            new_model.n_iter_ = 0
```
RomainGoussault, line 486: Why do we have to do that ?

RomainGoussault (changes requested):
Nice work @aygalic !

multi_partner_learning.py
bowni: Hmm, good question, with a quick search I didn't find anything about putting `n_iter` to `0`.

bowni: I ran test without it and can't see an impact. I commented it out for the moment (and asked @aygalic if he can clarify).

bowni: And finally I removed it after discussing with @aygalic . We should be all good

RomainGoussault (approved):
👍

PR #199, opened 2020-07-29
BugFix : scenario.py

aygalic:
A couple of bug were fixed in this commit

- Fixed a bug that throws an error and canceled every run

- Removed a log output : "Computation use the full dataset" would be displayed in loop at the start of every run.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -286,10 +286,10 @@ class Test_Scenario:
         def test_raiseException(self, create_Scenario):
             scenar = create_Scenario
             scenar.partners_list[0].x_train = scenar.dataset.x_train
-            with pytest.raises(AssertionError):
+            with pytest.raises(Exception):
                 if scenar.samples_split_option[0] == 'basic':
                     scenar.split_data()
-                else :
+                else:
                     scenar.split_data_advanced()
 
 
--- a/tests.py
+++ b/tests.py
@@ -282,14 +282,15 @@ class Test_Scenario:
                 scenar.instantiate_scenario_partners()
 
     class Test_split_data:
+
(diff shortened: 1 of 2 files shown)


bowni:
There is an issue in https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/18f45fa484170a09fc45494d1820eac8d2fed53a/tests.py#L287. It launches `split_data()` indifferently, whether the `samples_split_option` passed in the config file is `basic` or `advanced`. As in `main.py`, when it is `advanced` it should be `split_data_advanced()`.
@aygalic @natct10 @GauthierGar

natct10:
> There is an issue in
> 
> https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/18f45fa484170a09fc45494d1820eac8d2fed53a/tests.py#L287
> . It launches `split_data()` indifferently, whether the `samples_split_option` passed in the config file is `basic` or `advanced`. As in `main.py`, when it is `advanced` it should be `split_data_advanced()`.
> @aygalic @natct10 @GauthierGar

Bien vu

tests.py
```diff
@@ -284,7 +286,7 @@ def test_raiseException(self, create_Scenario):
             scenar = create_Scenario
             scenar.partners_list[0].x_train = scenar.dataset.x_train
             with pytest.raises(AssertionError):
-                if scenar.mpl.samples_split_option == 'advanced':
+                if scenar.samples_split_option == 'advanced':
                     scenar.split_data_advanced()
                 else :
```
natct10: ouh there is a space between `else` and `:`

PR #231, opened 2020-09-11
Catch fetch datasets errors

arthurPignet:
-create local copy of titanic when downloading it for the first time
-add try catch
-add retry, with a parameter in constants.py

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -5,7 +5,7 @@ Declaration of constants.
 
 # ML constants
 DEFAULT_BATCH_SIZE = 256
-MAX_BATCH_SIZE = 2**20
+MAX_BATCH_SIZE = 2 ** 20
 DEFAULT_GRADIENT_UPDATES_PER_PASS_COUNT = 8
 PATIENCE = 10  # patience for early stopping
 
@@ -30,3 +30,6 @@ ESC50 = "esc50"
 
 # Supported datasets
 SUPPORTED_DATASETS_NAMES = [MNIST, CIFAR10, TITANIC, ESC50]
+
+# Number of attempts allowed before raising an error while trying to download dataset
+NUMBER_OF_DOWNLOAD_ATTEMPTS = 3
--- a/datasets/dataset_cifar10.py
+++ b/datasets/dataset_cifar10.py
@@ -2,14 +2,20 @@
 """
 CIFAR10 dataset.
 """
+from time import sleep
+from urllib.error import HTTPError, URLError
+from loguru import logger
 
 from keras.datasets import cifar10
 from keras.models import Sequential
 from keras.layers import Dense, Dropout, Activation, Flatten
 from keras.layers import Conv2D, MaxPooling2D
-import keras
+from keras.optimizers import RMSprop
+from keras.utils import to_categorical
+
 from sklearn.model_selection import train_test_split
 
+import constants
 from . import dataset
 
 # Init dataset-specific variables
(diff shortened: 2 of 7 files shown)


constants.py
```diff
@@ -5,7 +5,7 @@
 
 # ML constants
 DEFAULT_BATCH_SIZE = 256
-MAX_BATCH_SIZE = 2**20
+MAX_BATCH_SIZE = 2 ** 20
```
bowni, line 8: Really that much?! It seems it is anterior to this PR but I raise the question here: why? cc @RomainGoussault @aygalic

datasets/dataset_cifar10.py
```diff
@@ -27,13 +33,35 @@ def preprocess_dataset_inputs(x):
...
+        except URLError as e:
+            logger.debug(
+                f'URL fetch failure on https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz : '
+                f'{e.errno} -- {e.reason}')
+            if attempts < constants.NUMBER_OF_DOWNLOAD_ATTEMPTS:
+                sleep(2)
+                attempts += 1
+            else:
+                raise
+
```
bowni, line 64: Nice 👍

bowni: Question : as the only thing that differs in handling the `HTTPError` and the `URLError` is the debug log message, can't it be factorized?

main.py
```diff
@@ -22,7 +22,7 @@
 import scenario
 import utils
 
-DEFAULT_CONFIG_FILE = "config.yml"
+DEFAULT_CONFIG_FILE = "config_quick_debug.yml"
```
bowni, line 25: I am not sure about that one

bowni (commented):
Great @arthurPignet, a couple of questions:

datasets/dataset_cifar10.py
```diff
@@ -27,13 +33,35 @@ def preprocess_dataset_inputs(x):
...
+                f'URL fetch failure on https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz : {e.code} -- {e.msg}')
+            if attempts < constants.NUMBER_OF_DOWNLOAD_ATTEMPTS:
+                sleep(2)
+                attempts += 1
+            else:
+                raise
+        except URLError as e:
+            logger.debug(
+                f'URL fetch failure on https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz : '
+                f'{e.errno} -- {e.reason}')
```
RomainGoussault, line 58: Why do we we log `{e.code} -- {e.msg}` for `HTTPError `and `{e.errno} -- {e.reason}` for the `URLError `?

constants.py
RomainGoussault: I guess for very big datasets we need to make sure the batch size is not greater than this `MAX_BATCH_SIZE`. Actually this makes me think that sometimes we clip the batch size so that is not greater than MAX_BATCH_SIZE but we don't give any feedback to the user.

RomainGoussault: I created an issue here https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/232

RomainGoussault (changes requested):
Nice work @arthurPignet !

datasets/dataset_cifar10.py
arthurPignet: In fact .msg and .reason are the same things for HTTPerror but msg does not exist for URLlib. The URLlib does not hold .code

arthurPignet: the .code will be for instance the famous 404, which I think is useful to log .

datasets/dataset_cifar10.py
arthurPignet: HTTP et URLError don't have the same attributs (even if .msg and .reason are basically the samefor HTTP, and HTTP have a .errno, but without interesting infos (None value)
I factorized that

datasets/dataset_titanic.py
```diff
@@ -72,9 +77,37 @@ def preprocess_dataset_inputs(x):
 
 def load_data():
     """Return a usable dataset"""
-
-    raw_dataset = pd.read_csv('https://web.stanford.edu/class/archive/cs/cs109/cs109.1166/stuff/titanic.csv',
-                              index_col=False)
+    path = Path(__file__).resolve().parents[0]
+    repertoire = str(path) + '/local_data/titanic/'
```
RomainGoussault, line 81: folder ?

datasets/dataset_titanic.py
```diff
@@ -72,9 +77,37 @@ def preprocess_dataset_inputs(x):
...
+    if not Path(repertoire).is_dir():
+        os.makedirs(repertoire)
+        os.chdir(repertoire)
```
RomainGoussault, line 84: You can use pathlib for this !

RomainGoussault (changes requested):

bowni (approved):

datasets/dataset_cifar10.py
bowni: Ok, clear and clean

RomainGoussault (approved):
Great work !

PR #238, opened 2020-09-21
Add imdb dataset v2

arthurPignet:
Completing the work of  @GauthierGar to add the IMDB dataset to the library
The branch is created from the branch improve-documentation, so the PR 235 needs to be merge before this one


bowni (changes requested):
Seems all good - well done @GauthierGar and @arthurPignet!
But... there is always a but otherwise it would be too easy 😃 : can you add in the `README` @arthurPignet that IMDb now joins the (not so) long list of pre-implemented datasets?

bowni (approved):
Great!

Thomas-Galtier (approved):

Issue #105, opened 2020-04-14
New contributivity method : 'federated step-by-step increments' 


bowni:
**Main idea** : measure the performance increments on the global validation dataset after every minibatch training.
When the whole federated training is done, average for every node the performance increments, this should give an estimate on how the model improved on every node, and thus might be an estimate of the contributivity of every node/dataset !

**Some technical considerations:**

- Increments from the first minibatches might be huge and not really relevant --> we should discard the X first increment. Or another option would be to only use the performance increments when the model has a performance of a least 50% (50 is an example).
- Increments from the last minibatches (just before the earlystopping) might be very noisy and therefore not relevant --> we could discard the X last increments.
- This approach might not be feasible when doing _sequential_ FL --> For instance, if we train node_1, then node_2, then node_3, it might be easier for node_1 to have a big increment because it starts first compared to node 3.

Note this approach can also be a cheap baseline for more advanced contributivity methods, in order to sample/discard specific nodes.

jeromechambost:
**In federated averaging computation:**

For each computation round:
- Performance of newly fit partner model on global test dataset is computed and stored for each partner
- Performance of aggregated model on global dataset is computed and stored

![Step-by-step performance increment](https://user-images.githubusercontent.com/58439042/81797162-b52e1b80-950e-11ea-9042-83becddb8a06.png)

At the end of computation, matrix of performance is stored with C rows (C = number of computation rounds) and N+1 columns (N = number of partners)

New matrix of performance increase is computed: For each computation round, increase in performance is calculated between newly fit model for each partner and previous aggregated model

Contributivity of partner is a function of the matrix of performance increase:
Contributivity funtion can be:
- Average of increase of performance
- Weighted average of increase of performance (increasing weights with increasing number of computation rounds)

Contributivity function takes 3 parameters:
- X the proportion of first computation rounds that are discarded due to high and uneven performance increase
- Y the proportion of last computation rounds that are discarded due to noise 

Further considerations
Performance is set to be accuracy but can be changed to another metric (if we choose the metric to be loss, we need to change calculation to adapt to the loss decrease instead of performance increase)

bowni:
I think we have some of what we need to move forward on this now that we merged PR #141.

At the end of a multi-partner computation, we have:

[...]

PR #239, opened 2020-09-22
Federated step by step contributivity methods

jeromechambost:
3 step by step contributivity methods


subtest/contributivity.py
```diff
@@ -938,6 +938,103 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
+
+        # Reshape scores matrices
+        scores_matrix_collective_reshape = np.reshape(score_matrix_collective_models,
+                                                      (epoch_count * minibatch_count))
+        score_matrix_per_partner_reshape = np.reshape(score_matrix_per_partner,
+                                                      (epoch_count * minibatch_count, partners_count))
+
+        # Calculate relative performance matrix
+        score_matrix_performance_rel = np.divide(score_matrix_per_partner_reshape,
+                                                 scores_matrix_collective_reshape[:, None])
```
Thomas-Galtier, line 1031: I might have misunderstood the method we wanted to implement, but I expected that here we would compute _the increment_ of the score for a partner on one minibatch step divided by _the increment_ of the collective score on one minibatch step. Here it seems that you manipulate the score at the end of each step and not its increase on each step. Was it your intention or is it a mistake?

Thomas-Galtier (changes requested):

subtest/contributivity.py
jeromechambost: @Thomas-Galtier see https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/105#issuecomment-648008117
After some tests it appeared that absolute increments were not optimal because the model could "unlearn" for each of the partners (meaning individual scores of each partner would be lower than the score of the aggregated model at previous step)

To solve the issue of regularization coming from this situation with @RomainGoussault , we had proposed to compute relative score increases (or decreases) rather than absolute, which is what is implemented here -- maybe this needs to be clarified in the issue as the method has slightly changed?

arthurPignet:
What results did you obtain when you tested it ? For instance I wonder if your scores are linked to the shapley values in anyway and if you are able to detect corrupted dataset

jeromechambost:
> What results did you obtain when you tested it ? For instance I wonder if your scores are linked to the shapley values in anyway and if you are able to detect corrupted dataset @arthurPignet 

Did not test it for corrupted dataset detection (I dont think this will be the best method for this but worth trying)
We tried some experiments where results were comparable to TMC Shapley but needing less computation time. We need to test it at large scale to properly compare the methods in all kinds of scenarios and identify those where results are close to shapley

main.py
```diff
@@ -20,7 +20,7 @@
 from subtest import scenario
 from subtest import utils
 
-DEFAULT_CONFIG_FILE = "./config.yml"
+DEFAULT_CONFIG_FILE = "./config_quick_debug.yml"
```
arthurPignet, line 23: I understood that it should remain ./config.yml, did I miss something ?

README.md
```diff
@@ -379,6 +382,16 @@ Example: `is_quick_demo: True`
 - [future prospect] **Federated step-by-step increments**:
 
   See open [issue #105](https://github.com/SubstraFoundation/distributed-learning-contributivity/issues/105).
+  
+  Measures the performance increments on the global validation dataset after every minibatch training - Gives an estimation on how the model improved on every node
+  Method suited only for federated averaging (for now)
```
arthurPignet, line 387: I did not see any kind of safegard for that, it will be nice to have one. Like an assertion, or a test, to check if we really use fedavg, and raise an error if not.  (maybe in the contributivity file, when you check for the method to compute

README.md
```diff
@@ -314,6 +314,9 @@ All methods available are:
 - "AIS_Kriging_S"
 - "SMCS"
 - "WR_SMC"
+- "Federated SBS linear"
+- "Federated SBS quadratic",
+- "Federated SBS constant"
```
arthurPignet, line 319: The README.md changed a lot, now this section belong to the documentation.md file.
Can you add it in the documentation.md ? 
-> mplc/documentation.md

arthurPignet (changes requested):

main.py
bowni: I agree @arthurPignet , I suppose @jeromechambost you did that to facilitate debugging right? 😃

README.md
bowni: Adding an `assert` there would be a good idea indeed @arthurPignet . @jeromechambost does that seem OK?

bowni:
@jeromechambost I think we are closing in here, well done! @arthurPignet pointed above a few "details", code-quality related. Do you think you can fix these things or do you want help on that?

README.md
jeromechambost: Indeed @arthurPignet 
Step by step method works and gives results for all learning approaches but is only relevant for federated averaging.
@bowni How about a test + warning rather than an assert so that the code can keep running when condition is not met ?

bowni: Good idea!

jeromechambost:
> @jeromechambost I think we are closing in here, well done! @arthurPignet pointed above a few "details", code-quality related. Do you think you can fix these things or do you want help on that?

arthurPignet (approved):
Great work !
Just a little thing, you left the default_config_file to config_quick_debug

bowni (changes requested):
Seems all good, except the last merge commit (I suppose it was for fixing conflicts with `master`?) where you seem @jeromechambost to have overwritten `README` with an old version (e.g. section `Using this library` is empty). Another symptom is that it 'unfixed' the setting of `DEFAULT_CONFIG_FILE` to `config.yml`.

mplc/contributivity.py
```diff
@@ -979,6 +1076,27 @@ def compute_contributivity(
...
+                               "averaging learning approach")
+            self.federated_SBS_linear(
+                current_scenario)
+        elif method_to_compute == "Federated SBS quadratic":
+            # Contributivity 11: step by step increments with quadratic importance increase
+            if current_scenario.multi_partner_learning_approach != "fedavg":
+                logger.warning("Step by step quadratic contributivity method is only suited for federated "
+                               "averaging learning approach")
+            self.federated_SBS_quadratic(
+                current_scenario)
```
natct10, line 1092: Could you please put this on one line?

mplc/contributivity.py
```diff
@@ -979,6 +1076,27 @@ def compute_contributivity(
...
+        elif method_to_compute == "Federated SBS constant":
+            # Contributivity 12: step by step increments with constant importance
+            if current_scenario.multi_partner_learning_approach != "fedavg":
+                logger.warning("Step by step constant contributivity method is only suited for federated "
+                               "averaging learning approach")
+            self.federated_SBS_constant(
+                current_scenario)
```
natct10, line 1099: Could you please put this on one line?

natct10 (commented):
Looks nice, thank you @jeromechambost!
And thank you for the comments in the code!!

bowni (commented):
@jeromechambost one thing I just realized, about the notebook.
In itself the notebook is interesting, so why not keep it in the repo. However, since `/experiments` is gitignored, the following

```python
root_folder = Path().absolute().parent / "experiments"
experiment_path = root_folder/ "test_score_matrices_2020-06-12_14h30_bis"
```

will become obsolete. So, what do you think, should we keep the notebook anyway, knowing that it is not meant to be executed, just to be read? Or would it make sense to pdf it and put it in another folder like `/studies`?
I don't know, this is thinking out loud!

jeromechambost:
> @jeromechambost one thing I just realized, about the notebook.
> In itself the notebook is interesting, so why not keep it in the repo. However, since `/experiments` is gitignored, the following
> 
> ```python
> root_folder = Path().absolute().parent / "experiments"
> experiment_path = root_folder/ "test_score_matrices_2020-06-12_14h30_bis"
> ```
> 
> will become obsolete. So, what do you think, should we keep the notebook anyway, knowing that it is not meant to be executed, just to be read? Or would it make sense to pdf it and put it in another folder like `/studies`?
> I don't know, this is thinking out loud!

@bowni Agree that it would become obsolete pretty fast and it was just for proof of concept of the methods anyway. I suggest to remove it for now.

bowni (approved):

PR #241, opened 2020-09-23
Add the PVRL method (rebased)

arthurPignet:
Following the (PR 180)[https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/180]

Need to be merged after the add imdb-dataset-V2

### DIFF
--- a/mplc/contributivity.py
+++ b/mplc/contributivity.py
@@ -963,12 +963,12 @@ class Contributivity:
         )
         mpl.compute_test_score()
         previous_loss = mpl.loss_collective_models[-1]
-        t = 0
-        while (t < the_scenario.epoch_count * the_scenario.partners_count or np.sum(
+        PVRL_epoch = 0
+        while (PVRL_epoch < the_scenario.epoch_count * the_scenario.partners_count or np.sum(
                 np.abs(partner_values - previous_partner_values)) / the_scenario.partners_count > epsilon):
-            t += 1
-            logger.info("t:", t)
-            logger.info("partner_values:", partner_values)
+            PVRL_epoch += 1
+            logger.info(f"t:{PVRL_epoch}")
+            logger.info(f"partner_values: {partner_values}")
             # Select the partner / the action
             is_partner_in = np.random.binomial(1, p=partner_values)
             while np.sum(is_partner_in) == 0:
--- a/mplc/contributivity.py
+++ b/mplc/contributivity.py
@@ -941,42 +941,46 @@ class Contributivity:
 
     # %% compute Partner value by reinforocment learning
 
-    def PVRL(self, the_scenario, learning_rate ):
+    def PVRL(self, the_scenario, learning_rate):
         start = timer()
         w = np.zeros(the_scenario.partners_count)
         partner_values = np.exp(w) / (1.0 + np.exp(w))
-        previous_partner_values  = np.zeros(the_scenario.partners_count)
+        previous_partner_values = np.zeros(the_scenario.partners_count)
         epsilon = 0.002
 
-        mpl= multi_partner_learning.MultiPartnerLearning(
-                    the_scenario.partners_list,
-                    1,
-                    the_scenario.minibatch_count,
-                    the_scenario.dataset,
-                    the_scenario.multi_partner_learning_approach,
-                    the_scenario.aggregation_weighting,
-                    is_early_stopping=False,
-                    is_save_data=True,
-                    save_folder=the_scenario.save_folder,
-                    init_model_from= "random_initialization",
-                    use_saved_weights=False,
-                    )
+        mpl = multi_partner_learning.MultiPartnerLearning(
+            the_scenario.partners_list,
+            1,
+            the_scenario.minibatch_count,
+            the_scenario.dataset,
+            the_scenario.multi_partner_learning_approach,
+            the_scenario.aggregation_weighting,
+            is_early_stopping=False,
+            is_save_data=True,
+            save_folder=the_scenario.save_folder,
+            init_model_from="random_initialization",
+            use_saved_weights=False,
+        )
         mpl.compute_test_score()
         previous_loss = mpl.loss_collective_models[-1]
-        t=0
-        while (t<the_scenario.epoch_count*the_scenario.partners_count or np.sum(np.abs(partner_values -previous_partner_values))/ the_scenario.partners_count > epsilon ):
-            t+=1
-            print("t:",t)
-            print("partner_values:",partner_values)
+        t = 0
+        while (t < the_scenario.epoch_count * the_scenario.partners_count or np.sum(
+                np.abs(partner_values - previous_partner_values)) / the_scenario.partners_count > epsilon):
+            t += 1
+            logger.info("t:", t)
+            logger.info("partner_values:", partner_values)
             # Select the partner / the action
             is_partner_in = np.random.binomial(1, p=partner_values)
             while np.sum(is_partner_in) == 0:
                 is_partner_in = np.random.binomial(1, p=partner_values)
 
             # apply one epoch with the selected partner to the previous model/ do the action
-            small_partner_list= [partner for partner, is_in in zip(the_scenario.partners_list,is_partner_in) if is_in == 1]
-            weights_folder = os.path.join(the_scenario.save_folder, 'model',the_scenario.dataset.name+'_final_weights.h5')
-            mpl= multi_partner_learning.MultiPartnerLearning(
+            small_partner_list = [partner for partner, is_in in zip(the_scenario.partners_list, is_partner_in) if
+                                  is_in == 1]
+            weights_folder = os.path.join(the_scenario.save_folder, 'model',
+                                          the_scenario.dataset.name + '_final_weights'
+                                                                      '.h5')
+            mpl = multi_partner_learning.MultiPartnerLearning(
                 small_partner_list,
                 1,
                 the_scenario.minibatch_count,
@@ -988,7 +992,7 @@ class Contributivity:
                 save_folder=the_scenario.save_folder,
                 init_model_from=weights_folder,
                 use_saved_weights=True,
-                )
+            )
             mpl.compute_test_score()
             loss = mpl.loss_collective_models[-1]
 
@@ -998,8 +1002,10 @@ class Contributivity:
             # Update the weight according to the REINFORCE method
             new_w = np.zeros(the_scenario.partners_count)
             for i in range(the_scenario.partners_count):
-                new_w[i] = w[i] + learning_rate * G * dp_dw[i] * (is_partner_in[i] / partner_values[i]  - (1.0 - is_partner_in[i])  / (1.0 -partner_values[i])  - prodp / (1.0-prodp) / (1.0 -partner_values[i]) )
-            w=new_w
+                grad = is_partner_in[i] / partner_values[i] - (1.0 - is_partner_in[i]) / (
+                        1.0 - partner_values[i]) - prodp / (1.0 - prodp) / (1.0 - partner_values[i])
+                new_w[i] = w[i] + learning_rate * G * dp_dw[i] * grad
+            w = new_w
             # Update values before the next round
             previous_partner_values = partner_values
             partner_values = np.exp(w) / (1.0 + np.exp(w))
--- a/mplc/contributivity.py
+++ b/mplc/contributivity.py
@@ -939,7 +939,7 @@ class Contributivity:
             end = timer()
             self.computation_time_sec = end - start
 
-    # %% compute Partner value by reinforocment learning
+    # %% compute Partner value by reinforcement learning
 
     def PVRL(self, the_scenario, learning_rate):
         start = timer()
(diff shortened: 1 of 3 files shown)


Thomas-Galtier (approved):

mplc/contributivity.py
```diff
@@ -938,6 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
+                        1.0 - partner_values[i]) - prodp / (1.0 - prodp) / (1.0 - partner_values[i])
+                new_w[i] = w[i] + learning_rate * G * dp_dw[i] * grad
+            w = new_w
+            # Update values before the next round
+            previous_partner_values = partner_values
+            partner_values = np.exp(w) / (1.0 + np.exp(w))
+            previous_loss = loss
+
+        self.name = "PVRL"
+        print(f"PVRL: the final accuracy is  {mpl.test_score}")
```
natct10, line 1016: It would be nice to log this instead of printing it

mplc/contributivity.py
```diff
@@ -938,6 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
+            use_saved_weights=False,
+        )
+        mpl.compute_test_score()
+        previous_loss = mpl.loss_collective_models[-1]
+        t = 0
+        while (t < the_scenario.epoch_count * the_scenario.partners_count or np.sum(
+                np.abs(partner_values - previous_partner_values)) / the_scenario.partners_count > epsilon):
+            t += 1
+            print("t:", t)
+            print("partner_values:", partner_values)
```
natct10, line 972: log it?

mplc/multi_partner_learning.py
```diff
@@ -554,21 +553,10 @@ def init_with_agg_models(self):
 
         self.prepare_aggregation_weights()
         partners_model_list = []
-        for _ in self.partners_list:
+        for partner in self.partners_list:
```
natct10, line 556: +1

config_quick_debug.yml
```diff
@@ -10,14 +10,12 @@ scenario_params_list:
...
      - ['basic', 'random']
-     - ['basic', 'stratified']
-     - ['advanced', [[4, 'shared'], [6, 'shared'], [4, 'specific']]]
    multi_partner_learning_approach:
     - 'fedavg'
    aggregation_weighting:
     - 'uniform'
    methods:
-    - ["Shapley values", "Independent scores"]
+    - ["TMCS", "PVRL"]
```
natct10, line 18: Is this file supposed to evolve?

natct10 (commented):
Looks cool, thank you @arthurPignet!

config_quick_debug.yml
Thomas-Galtier: We said thuesday that  the config_debug can change now, as we will create a config_travis. I may have anticipated this change here. I will put config_debug back as it was.

mplc/multi_partner_learning.py
```diff
@@ -208,14 +209,14 @@ def compute_test_score(self):
         self.test_score = model_evaluation_test_data[1]  # 0 is for the loss
         self.nb_epochs_done = self.epoch_index + 1
 
-        # Plot training history
-        # TODO: move the data saving and plotting in dedicated functions
+        # Plot training history # TODO: move the data saving and plotting in dedicated functions
```
RomainGoussault, line 212: I think this TODO is done, we can remove it

mplc/contributivity.py
```diff
@@ -938,6 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
             end = timer()
             self.computation_time_sec = end - start
 
+    # %% compute Partner value by reinforocment learning
+
+    def PVRL(self, the_scenario, learning_rate):
```
RomainGoussault, line 945: Can we provide the link to the PVRL paper ?

mplc/multi_partner_learning.py
```diff
@@ -109,7 +109,8 @@ def compute_test_score_for_single_partner(self, partner):
 
         # Save model score on test data
         self.test_score = model_evaluation_test_data[1]  # 0 is for the loss
-        self.nb_epochs_done = (es.stopped_epoch + 1) if es.stopped_epoch != 0 else self.epoch_count
+        self.loss_collective_models.append(model_evaluation_test_data[0])  # store the loss for PVRL
+        self.nb_epochs_done = (es.stopped_epoch + 1) if self.is_early_stopping else self.epoch_count
```
RomainGoussault, line 113: Looking at Eric's comment here --> https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/141#discussion_r428158162
I think this won't work.

mplc/contributivity.py
```diff
@@ -7,6 +7,7 @@
 
 import bisect
 import datetime
+import os
```
RomainGoussault, line 10: Import pathlib ;)

RomainGoussault (changes requested):
Nice work !
Can we provide a test for this contributivity method ? An end to end test might be too long to run so maybe we could the PVRL method only ?

mplc/contributivity.py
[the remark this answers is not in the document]
Thomas-Galtier: There is no PVRL paper yet but only a paper on DVRL

mplc/multi_partner_learning.py
Thomas-Galtier: Actually this line was not working when the earlystooping was deactivated because the `es` object was no more instantiated in that case. This new line also to fix this.

mplc/contributivity.py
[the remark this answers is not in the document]
RomainGoussault: Then maybe add one or two lines describing the method. Also we should update the doc here: https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/master/mplc/docs/documentation.md#configuration-of-contributivity-measurement-methods-to-be-tested

mplc/multi_partner_learning.py
RomainGoussault: Ok I understand.
However I think Eric's [comment](https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/141#discussion_r428158162) still holds: 

> Via the callback the EarlyStopping provides stopped_epoch. But the way it is implemented, if early stopping wasn't trigger, this returns 0. 


So if `is_early_stopping = true` but the early stopping was never triggered, `stopped_epoch = 0`

mplc/contributivity.py
```diff
@@ -938,6 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
             end = timer()
             self.computation_time_sec = end - start
 
+    # %% compute Partner value by reinforocment learning
```
natct10, line 943:
```suggestion
    # %% compute Partner value by reinforcement learning
```

natct10: Is there a meaning for `%%`?

arthurPignet: I don't know 
Maybe in some IDE ? All the comments about the contributivity measures follow this format

mplc/docs/documentation.md
```diff
@@ -403,7 +338,7 @@ dataset = dataset.Dataset(
 ```
 
 ### Model generator
-This function provides the model which will be trained by the scenario object. Currently the library handles compiled Keras' model (see MNIST, ESC50, IMDB and CIFAR10 datasets), and Scikit-Learn Linear Regression (see the TITANIC dataset).  
+This function provides the model use, which will be trained by the scenario object.
```
bowni, line 341: This is weird, it looks like `documentation.md` goes back in time versus its current state in `master`?

mplc/contributivity.py
```diff
@@ -938,103 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
             end = timer()
             self.computation_time_sec = end - start
 
-    def federated_SBS_linear(self, the_scenario):
-        start = timer()
-        logger.info(
-            "# Launching computation of perf. scores of linear "
-            "performance increase compared to previous collective model")
-
-        relative_perf_matrix = self.compute_relative_perf_matrix(the_scenario)
```
bowni, line 947: Same in this file, it looks like it wasn't well rebased, and it erases quite a lot of recently merged material.

mplc/contributivity.py
bowni: I concur with @natct10 remark 😄 Any reason @arthurPignet @Thomas-Galtier why it was dismissed?

bowni (changes requested):

mplc/docs/documentation.md
natct10: I think this PR will need to be rebased as you pointed before @bowni

mplc/multi_partner_learning.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
natct10, line 1: Why remove the encoding tag? I'm curious

mplc/multi_partner_learning.py
```diff
@@ -102,14 +101,19 @@ def compute_test_score_for_single_partner(self, partner):
...
         # Evaluate trained model
         model_evaluation_test_data = self.evaluate_model(model, self.test_data)
         logger.info(f"   Model evaluation on test data: "
                     f"{list(zip(model.metrics_names, ['%.3f' % elem for elem in model_evaluation_test_data]))}")
 
         # Save model score on test data
         self.test_score = model_evaluation_test_data[1]  # 0 is for the loss
-        self.nb_epochs_done = (es.stopped_epoch + 1) if es.stopped_epoch != 0 else self.epoch_count
+        self.nb_epochs_done = (es.stopped_epoch + 1) if (self.is_early_stopping
+                                                         and es.stopped_epoch != 0) else self.epoch_count
```
RomainGoussault, line 116: cf my previous comment:
https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/241#discussion_r499772512

RomainGoussault (changes requested):

mplc/contributivity.py
```diff
@@ -938,103 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
-        # Return contributivity scores
-        self.name = "Federated step by step constant scores"
-        self.contributivity_scores = contributivity_scores
+        mpl.compute_test_score()
+        previous_loss = mpl.loss_collective_models[-1]
+        t = 0
+        while (t < the_scenario.epoch_count * the_scenario.partners_count or np.sum(
+                np.abs(partner_values - previous_partner_values)) / the_scenario.partners_count > epsilon):
+            t += 1
+            logger.info("t:", t)
```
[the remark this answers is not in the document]
RomainGoussault: Actually this only logs the first "t".
We should do something like:             `logger.info(f"t: {t}")`

mplc/contributivity.py
```diff
@@ -938,103 +940,88 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
+            logger.info("partner_values:", partner_values)
```
RomainGoussault, line 972: same here, it will only log the first string

mplc/multi_partner_learning.py
arthurPignet: I think I missed something
In the case of there is an early stopping, and it was triggered (ie es.stopped_epoch > 0), the number of epoch done is equal to es.stopped_epoch right ?
At the opposite, if there is no es, or if it was not triggered, the number of epoch done should be epoch_count ?

mplc/contributivity.py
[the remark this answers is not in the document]
arthurPignet: In fact this t represents the number of epoch done. I renamed it PVRL_epoch. Due to the implementation of PVRL, most of the log will make no sense. (as a new mpl object is recreated at each epoch). It will be solved with the refacto of mpl, but I need changes which are done in the refacto to modify that.

bowni (approved):

mplc/contributivity.py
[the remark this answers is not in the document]
bowni: There are a few words about it in the `documentation.md`

mplc/docs/documentation.md
```diff
@@ -350,13 +350,13 @@ The methods are detailed below:
   - `["SMCS"]` **Stratified Monte Carlo Shapley with replacement**
   - `["WR_SMC"]` **Stratified Monte Carlo Shapley without replacement**
 
-- [In progress]  **Partner Valuation by Reinforcement Learning** :
+- **Partner Valuation by Reinforcement Learning** :
```
natct10, line 353:
```suggestion
- **Partner Valuation by Reinforcement Learning**:
```

natct10 (approved):
Just one suggestion and approved!

bowni (approved):

RomainGoussault (approved):
🎉 !!

PR #247, opened 2020-09-29
Add partial corruption

Thomas-Galtier:
Corrupt only a part of a partner's data

### DIFF
--- a/mplc/partner.py
+++ b/mplc/partner.py
@@ -33,18 +33,16 @@ class Partner:
         self.y_test = None
 
     def corrupt_labels(self, proportion_corrupted):
-        if proportion_corrupted>1 or proportion_corrupted<0:
+        if not 0 < proportion_corrupted < 1:
             raise ValueError(
-                "The proportion of labells to corrupted ["
-                + str(proportion_corrupted)
-                + "] must be between 0 and 1."
+                f"The proportion of labels to corrupted  {proportion_corrupted} must be between 0 and 1."
             )
-        
-        # Select the indices where the label will be off-set 
-        n = int(len(self.y_train)*proportion_corrupted)
+
+        # Select the indices where the label will be off-set
+        n = int(len(self.y_train) * proportion_corrupted)
         print()
         idx = sample(list(range(len(self.y_train))), n)
-        
+
         # Off-set  the labels
         for label in self.y_train[idx]:
             idx_max = np.argmax(label)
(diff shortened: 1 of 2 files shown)


mplc/partner.py
```diff
@@ -32,12 +32,44 @@ def __init__(self, partner_id):
         self.y_val = None
         self.y_test = None
 
-    def corrupt_labels(self):
-        for label in self.y_train:
+    def corrupt_labels(self, proportion_corrupted):
+        if proportion_corrupted>1 or proportion_corrupted<0:
```
natct10, line 36: I think this could be easier to read as `if not 0 < proportion_corrupted < 1:`.
And you could also use f-strings for the exception error.
But this is only a suggestion

mplc/partner.py
```diff
@@ -32,12 +32,44 @@ def __init__(self, partner_id):
...
+        for label in self.y_train[idx]:
             idx_max = np.argmax(label)
             label[idx_max] = 0.0
             label[idx_max - 1] = 1.0
 
-    def shuffle_labels(self):
-        for label in self.y_train:
-            label = shuffle(label)
+    def shuffle_labels(self, proportion_shuffled):
+        if proportion_shuffled>1 or proportion_shuffled<0:
```
natct10, line 55: I'd make the same suggestion here :D

mplc/scenario.py
```diff
@@ -58,9 +58,11 @@ def __init__(self,
         :param samples_split_option: ['basic', 'random'] (default),
                                      ['basic', 'stratified']
                                      or ['advanced', [[nb of clusters (int), 'shared' or 'specific']]].
-        :param corrupted_datasets: ['not_corrupted' (default), 'shuffled' or 'corrupted'].
+        :param corrupted_datasets: ['not_corrupted' (default), [ 'shuffled', proportion_suffled] or ['corrupted', proportion_corrupted]].
```
natct10, line 61:
```suggestion
        :param corrupted_datasets: ['not_corrupted' (default), ['shuffled', proportion_suffled] or ['corrupted', proportion_corrupted]].
```

mplc/scenario.py
```diff
@@ -163,7 +165,7 @@ def __init__(self,
         if corrupted_datasets is not None:
             self.corrupted_datasets = corrupted_datasets
         else:
-            self.corrupted_datasets = ["not_corrupted"] * self.partners_count  # default
+            self.corrupted_datasets = ["not_corrupted" ] * self.partners_count  # default
```
natct10, line 168:
```suggestion
            self.corrupted_datasets = ["not_corrupted"] * self.partners_count  # default
```

natct10 (commented):
I didn't test yet, but it looks cool!

mplc/partner.py
```diff
@@ -32,12 +31,39 @@ def __init__(self, partner_id):
...
+
+        # Select the indices where the label will be off-set
+        n = int(len(self.y_train) * proportion_corrupted)
+        idx = sample(list(range(len(self.y_train))), n)
+
+        # Off-set  the labels
+        for label in self.y_train[idx]:
             idx_max = np.argmax(label)
             label[idx_max] = 0.0
             label[idx_max - 1] = 1.0
```
bowni, line 48: So here the hypothesis is that labels will always be `0` or `1`. @arthurPignet is that coherent / robust with regard to how you refactored the `Dataset` object and the way it is supposed to be used?

bowni (approved):

bowni:
@Thomas-Galtier it looks like Travis and Flake are unhappy with 5 little things 😄 

```
./mplc/scenario.py:73:117: W291 trailing whitespace
./mplc/scenario.py:462:43: E203 whitespace before ':'
./mplc/scenario.py:465:39: E203 whitespace before ':'
./mplc/scenario.py:531:32: E203 whitespace before ':'
./mplc/scenario.py:536:32: E203 whitespace before ':'
```

arthurPignet:
I think it would be great if you could add a default value (1 for instance) to this parameter. The old use case would be robust, and it is maybe easier for a new user. 
Otherwise, the example notebooks need to be updated

arthurPignet (approved):

natct10 (approved):
That's a good idea, thank you @Thomas-Galtier!

PR #255, opened 2020-10-16
Mpl refacto

arthurPignet:
The final objective is to refacto the mpl methods into classes. The use of the library will be clearer, and add a new method will be easier. 

Don't hesitate to comment ! This is a major refacto, with many different possible solutions.


mplc/constants.py
```diff
@@ -50,9 +51,9 @@
...
-    "fedavg",
-    "seq-pure",
-    "seq-with-final-agg",
-    "seqavg",
-]
+MULTI_PARTNER_LEARNING_APPROACHES = {
+    "fedavg": multi_partner_learning.FederatedAverageLearning,
+    "seq-pure": multi_partner_learning.SequentialLearning,
+    "seq-with-final-agg": multi_partner_learning.SequentialWithFinalAggLearning,
+    "seqavg": multi_partner_learning.SequentialAverageLearning
```
RomainGoussault, line 58: We should stick with snake case here I think.

arthurPignet: For the keys or the values ? The values are classes, so I used CamelCase. And for the keys I copied the current parameters

bowni: - For class names PEP8 recommends CapWords, so that is good.
- For the names of the multi-partner learning approaches, I think @RomainGoussault you mean use underscores instead of dashes, right? I don't know. I liked the idea of marking the differences between variables and functions names (snake_case), and "actual names".

CONTRIBUTING.md
```diff
@@ -163,6 +163,22 @@ Notebooks now have really efficient ways to share your code and display your out
 - [Binder](https://mybinder.org/) that lets you turn a Git repository into a collection of interactive notebooks!
 - [Voila](https://github.com/voila-dashboards/voila) is a newcomer that will allow to build dashboads for your presentations!
 
+#### 2.6 Release a new build 
+
+To release a new version on Pypi, go at the root of the repository, and trigger the build with `pip`.
```
bowni, line 168: Not sure what "trigger the build with `pip`" means: should the user run a `pip` command he is supposed to know of already?

README.md
```diff
@@ -46,6 +46,8 @@ The objective is to compare the contributivity figures obtained with the differe
 
 The public datasets currently supported are: MNIST, CIFAR10, TITANIC, ESC50 and IMDB.
 
+
+
 ### Structure of the library
```
bowni, line 51: There seems to be 2 unnecessary blank lines there
```suggestion
The public datasets currently supported are: MNIST, CIFAR10, TITANIC, ESC50 and IMDB.

### Structure of the library
```

README.md
```diff
@@ -60,11 +62,33 @@ From the start of this work it seemed very useful to be able to simulate differe
...
+
+### Installation 
+
+##### Using pip
+
+```bash
+$ pip install mplc
+```
+
+This installs the last packaged version on pypi. If the installation fails on your platform, please open an issue to let us know.
```
bowni, line 79:
```suggestion
This installs the last packaged version on PyPI. If the installation fails on your platform, please open an issue to let us know.
```

mplc/datasets/dataset_titanic.py
```diff
@@ -141,3 +143,74 @@ def train_val_split_global(x, y):
...
+        else:
+            loss = log_loss(y_eval, self.predict(x_eval))  # mimic keras model evaluation
+            accuracy = self.score(x_eval, y_eval)
+            model_evaluation = [loss, accuracy]
+
+        return model_evaluation
+
+    def save_weight(self, path):
+        if self.coef_ is None:
+            raise ValueError('The model has never been fit, coef and intercept are set to None')
```
bowni, line 185: I would soften the message, like: "Coef and intercept are set to None, it seems the model has not been fit properly."
```suggestion
    def save_weight(self, path):
        if self.coef_ is None:
            raise ValueError('Coef and intercept are set to None, it seems the model has not been fit properly.')
```

mplc/datasets/dataset_titanic.py
```diff
@@ -141,3 +143,74 @@ def train_val_split_global(x, y):
...
+            logger.debug('Automatically switch file format from .h5 to .npy')
+            path.replace('.h5', '.npy')
+        np.save(path, self.get_weights())
+
+    def load_weights(self, path):
+        if '.h5' in path:
+            logger.debug('Automatically switch file format from .h5 to .npy')
+            path.replace('.h5', '.npy')
+        weights = load(path)
+        self.set_weight(weights)
```
bowni, line 196: Seems a bit too hacky to me. I can't remember, what is the point of switching extensions again?
If it applies only to models created from this class, where else in the code would it be a problem to have `.npy` files?

mplc/docs/documentation.md
```diff
@@ -50,12 +51,12 @@ To run a multi-partner learning and contributivity measurement experiment, you h
...
 - what contributivity measurement approach(es) will be run
 
-There are only 2 mandatory parameters to define a scenario: `partners_count` and `amounts_per_partner`. Many more exists but are optional as default values are configured. You can browse them all in below section [Scenario parameters](#scenario-parameters).
+There are only 2 mandatory parameters to define a scenario: `partners_count` and `amounts_per_partner`. Many more exist but are optional as default values are configured. You can browse them all in below section [Scenario parameters](#scenario-parameters).
 
 For this very first scenario, you could for example want to see what is happening with 3 partners, where the first one gets 20% of the total dataset, the second one 50% and the third one 30% (for a total of 100%!):
 
 ```python
-from subtest.scenario import Scenario
+from mplc.scenario import Scenario
```
bowni, line 59: Good catch!

mplc/partner.py
```diff
@@ -41,3 +41,50 @@ def corrupt_labels(self):
...
+    def __init__(self, partner, mpl):
+        """
+        :type partner: Partner
+        :type mpl: MultiPartnerLearning
+        """
+        self.mpl = mpl
+        self.id = partner.id
+        self.batch_size = partner.batch_size
+        self.minibatch_count = mpl.minibatch_count
+        self.partner_parent = partner
```
bowni, line 57: For coherence and readibility shouldn't we call the argument `parent` and the attribute `self.partner_parent` with the same name instead of 2 different ones?

mplc/partner.py
```diff
@@ -41,3 +41,50 @@ def corrupt_labels(self):
...
+        self.model_weight = None
+
+        self.minibatched_x_train = np.nan * np.zeros((self.minibatch_count, self.batch_size))
+        self.minibatched_y_train = np.nan * np.zeros((self.minibatch_count, self.batch_size))
```
bowni, line 61: I don't get why there is a second dimension of size `self.batch_size`

mplc/scenario.py
```diff
@@ -229,7 +232,7 @@ def __init__(self,
         # Miscellaneous
         # -------------
 
-        # Scenario id and number of repetition
+        # Scenario partner_id and number of repetition
 
         self.scenario_id = scenario_id
         self.n_repeat = repeats_count
```
bowni, line 238: Looks like a side effect from a search & replace?

bowni (commented):
First review, part 1/2 😃

mplc/datasets/dataset_titanic.py
arthurPignet: Yes it is cleary a hack. 
The problem is that we use everywhere in the code filename ending with .h5, as the code is written to deal with Keras' model. 
We can maybe switch the code to `.npy`, that could solve the problem for the weights, but not for the full model.

mplc/partner.py
arthurPignet: I missunderstood the role of self.batch_size. I remove this second dimension

mplc/multi_partner_learning.py
```diff
@@ -59,592 +56,474 @@ def __init__(self,
...
         # After last epoch or if early stopping was triggered, evaluate model on the global testset
-        logger.info("### Evaluating model on test data:")
-        model_evaluation_test_data = self.evaluate_model(model_to_evaluate, self.test_data)
-        logger.info(f"   Model metrics names: {model_to_evaluate.metrics_names}")
-        logger.info(f"   Model metrics values: {['%.3f' % elem for elem in model_evaluation_test_data]}")
-        self.test_score = model_evaluation_test_data[1]  # 0 is for the loss
-        self.nb_epochs_done = self.epoch_index + 1
+        self.history.log_final_model_perf()
+        # Save the model's weights
+        self.save_final_model_weights(self.get_model())
```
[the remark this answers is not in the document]
arthurPignet: Yes you are right, I will change that, we should manipulate only weights I think

mplc/constants.py
natct10: I'd rather go for snake case, but keeping method names like this is ok for me

setup.py
```diff
@@ -8,7 +8,7 @@
 
 setup(
     name="mplc",
-    version='0.0.0.1',
+    version='0.1',
     author="SubstraFoundation",
     author_email="contact@substra.org",
     description="A distributed learning contributivity package",
```
natct10, line 14: What about:
```suggestion
    description="A distributed learning contributivity measure package",
```
?

mplc/scenario.py
```diff
@@ -172,19 +172,22 @@ def __init__(self,
...
         # Define how federated learning aggregation steps are weighted. Toggle between 'uniform' and 'data_volume'
         # Default is 'uniform'
-        if aggregation_weighting in ['uniform', 'data_volume']:
-            self.aggregation_weighting = aggregation_weighting
-        else:
-            raise ValueError(f"aggregation_weighting approach'{self.aggregation_weighting}' is not a valid approach. ")
+        try:
+            self.aggregation = constants.AGGREGATORS[aggregation_weighting]
+        except KeyError:
+            raise ValueError(f"aggregation approach'{aggregation_weighting}' is not a valid approach. ")
```
natct10, line 190:
```suggestion
            raise ValueError(f"aggregation approach '{aggregation_weighting}' is not a valid approach. ")
```

mplc/partner.py
```diff
@@ -40,4 +40,51 @@ def corrupt_labels(self):
 
     def shuffle_labels(self):
         for label in self.y_train:
-            label = shuffle(label)
+            shuffle(label)
+
+
+class PartnerMpl:
+
```
natct10, line 47:
```suggestion
```

natct10 (commented):
This is a really nice refacto, well done & thank you @arthurPignet!
I am still checking this, but it looks neat!

mplc/contributivity.py
```diff
@@ -156,7 +154,7 @@ def compute_SV(self, the_scenario):
         characteristic_function = []
 
         for coalition in coalitions:
-            characteristic_function.append(self.not_twice_characteristic(coalition, the_scenario))
+            characteristic_function.append(self.not_twice_characteristic(coalition, self.scenario))
```
natct10, line 157: Not sure to follow here why `not_twice_characteristic` is taking a second arg whereas the method expect only one?
`def not_twice_characteristic(self, subset)`. Am I missing something?

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        plt.legend()
+        plt.ylim([0, 1])
+        plt.savefig(self.save_folder / "graphs/all_partners.png")
+        plt.close()
+
+
+class Aggregator(ABC):
+    def __init__(self, mpl):
+        """
+
```
natct10, line 124:
```suggestion
```

setup.py
arthurPignet: Yeah the description needs to be changed I approve

mplc/contributivity.py
arthurPignet: No you are right I missed two calls to this function while refacting

mplc/multi_partner_learning.py
```diff
@@ -5,41 +5,34 @@
...
-from sklearn.linear_model import LogisticRegression
-from sklearn.metrics import log_loss
 
 from . import constants
+from .mpl_utils import History, UniformAggregator
+from .partner import PartnerMpl
 
 
-class MultiPartnerLearning:
+class MultiPartnerLearning(ABC):
```
natct10, line 21: Neat

mplc/partner.py
```diff
@@ -40,4 +40,51 @@ def corrupt_labels(self):
 
     def shuffle_labels(self):
         for label in self.y_train:
-            label = shuffle(label)
+            shuffle(label)
+
+
+class PartnerMpl:
```
natct10, line 46: Any chance to rename this one?

arthurPignet: Yes please ! I am not satisfied either, but I was not able to find a more suitable name. Any suggestion ?

natct10: `PartnerInMlp`...? I'm not super inspired...

mplc/contributivity.py
```diff
@@ -940,93 +949,86 @@ def without_replacment_SMC(self, the_scenario, sv_accuracy=0.01, alpha=0.95):
...
+                                      batch_size=constants.DEFAULT_BATCH_SIZE,
+                                      verbose=0,
+                                      )
+        previous_loss = hist[0]
+        while (mpl.epoch_index < mpl.epoch_count):
+            # or (np.sum(np.abs(partner_values - previous_partner_values)) / self.scenario.partners_count > epsilon):
+
+            # Select the partners / the action
+            mpl.partners_list = []
+            while mpl.partners_count <= 1:
```
Thomas-Galtier, line 982:
```suggestion
            while mpl.partners_count ==0:
```

mplc/multi_partner_learning.py
```diff
@@ -5,42 +5,34 @@
...
-
+class MultiPartnerLearning(ABC):
     def __init__(self,
                  partners_list,
                  epoch_count,
                  minibatch_count,
                  dataset,
-                 multi_partner_learning_approach,
-                 aggregation_weighting="uniform",
+                 aggregation=UniformAggregator,
```
Thomas-Galtier, line 28: Can we change the default to data_volume?

bowni: Ok for me

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        self.mpl = mpl
+        self.save_folder = mpl.save_folder
+        self.nb_epochs_done = 0
+        self.score = None  # Final score evaluated on the test dataset at the end of the training
+        self.metrics = ['val_accuracy', 'val_loss', 'loss', 'accuracy']
+        tab = {key: np.nan * np.zeros((mpl.epoch_count, mpl.minibatch_count)) for key in self.metrics}
+        self.history = {partner.id: deepcopy(tab) for partner in mpl.partners_list}
+        self.history['model'] = {'val_accuracy': np.zeros((mpl.epoch_count, mpl.minibatch_count)),
+                                 'val_loss': np.zeros((mpl.epoch_count, mpl.minibatch_count))}
```
bowni, line 30: Maybe for readibility you could rename @arthurPignet the key `'model'` into `'multi_partner_model`'?

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def log_model_val_perf(self):
+        model = self.mpl.build_model()
+        hist = model.evaluate(self.mpl.val_data[0],
+                              self.mpl.val_data[1],
+                              batch_size=constants.DEFAULT_BATCH_SIZE,
+                              verbose=0,
+                              )
```
bowni, line 38: During the first review I was a bit hesitant with putting the actual evaluation task in the `History` object. I still am, but let's keep it like this and see how it goes. If needed we should be able to move it elsewhere without too much trouble

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        self.history['model']['val_loss'][self.mpl.epoch_index, self.mpl.minibatch_index] = hist[0]
+        self.history['model']['val_accuracy'][self.mpl.epoch_index, self.mpl.minibatch_index] = hist[1]
+
+        if self.mpl.minibatch_index >= self.mpl.minibatch_count - 1:
```
bowni, line 42: Shouldn't it be a `== 1` test? `>` should probably raise an error

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        hist = model.evaluate(self.mpl.test_data[0],
+                              self.mpl.test_data[1],
+                              batch_size=constants.DEFAULT_BATCH_SIZE,
+                              verbose=0,
+                              )
+        self.score = hist[1]
+        self.nb_epochs_done = self.mpl.epoch_index + 1
+        logger.info(f"   Model metrics names: {self.mpl.metrics_names}")
+        logger.info(f"   Model metrics values: {['%.3f' % elem for elem in hist]}")
+
```
bowni, line 69: It would be a bit easier to read if that method was next to `log_model_val_perf()`

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        self.history['model']['val_loss'][self.mpl.epoch_index, self.mpl.minibatch_index] = hist[0]
+        self.history['model']['val_accuracy'][self.mpl.epoch_index, self.mpl.minibatch_index] = hist[1]
+
+        if self.mpl.minibatch_index >= self.mpl.minibatch_count - 1:
+            logger.info(f"   Model evaluation at the end of the epoch: "
+                        f"{['%.3f' % elem for elem in hist]}")
+
+    def log_partner_perf(self, partner_id, partner_index, history):
+        for key in self.metrics:
+            self.history[partner_id][key][self.mpl.epoch_index, self.mpl.minibatch_index] = history[key][-1]
```
Thomas-Galtier, line 48: When running a seq-pure scenario, I got a bug on this line saying the History object is not subscriptable

mplc/multi_partner_learning.py
```diff
@@ -60,594 +53,427 @@ def __init__(self,
...
 
-            # Split the train dataset in mini-batches
-            self.split_in_minibatches()
+    def early_stop(self):
+        logger.debug("      Checking if early stopping criteria are met:")
+        if self.is_early_stopping:
+            # Early stopping parameters
+            if (
+                    self.epoch_index >= constants.PATIENCE
+                    and self.history.history['loss'][-1] > self.history.history['loss'][-constants.PATIENCE]
```
Thomas-Galtier, line 122: I get a `key error : 'loss'` as soon as `self.epoch_index >= constants.PATIENCE`

mplc/mpl_utils.py
```diff
@@ -0,0 +1,168 @@
...
+                              self.mpl.test_data[1],
+                              batch_size=constants.DEFAULT_BATCH_SIZE,
+                              verbose=0,
+                              )
+        self.score = hist[1]
+        self.nb_epochs_done = self.mpl.epoch_index + 1
+        logger.info(f"   Model metrics names: {self.mpl.metrics_names}")
+        logger.info(f"   Model metrics values: {['%.3f' % elem for elem in hist]}")
+
+    def partners_to_dataframe(self):
```
bowni, line 70: Sorry if it is a false alert: I can't find this method used anywhere else - what is it made for?

mplc/partner.py
```diff
@@ -70,3 +70,49 @@ def shuffle_labels(self, proportion_shuffled):
...
+        # Shuffle the dataset
+        idx = np.random.permutation(len(self.partner_parent.x_train))
+        x_train, y_train = self.partner_parent.x_train[idx], self.partner_parent.y_train[idx]
+
+        # Split the samples and labels
+        self.minibatched_x_train = np.split(x_train, (split_indices[:-1] * len(x_train)).astype(int))
+        self.minibatched_y_train = np.split(y_train, (split_indices[:-1] * len(y_train)).astype(int))
+
+    def build_model(self):
+        return self.mpl.build_model_from_weights(self.model_weights)
```
bowni, line 118: Just a question I just asked myself: is it OK from a python perspective / best practices, to have a method using `self.model_weights` as principal input, but no method actually updating `self.model_weights` from `None` (as these are in other objects)?

bowni (commented):
First part of my review! I still have `contributivity.py` and `multi_partner_learning.py` to review tomorrow 😃

mplc/contributivity.py
```diff
@@ -1084,16 +1086,21 @@ def federated_SBS_constant(self, the_scenario):
...
         # Define proportion of initial and final computation rounds to skip (default 10% each)
         init_comp_rounds_skipped = 0.1
         final_comp_rounds_skipped = 0.1
 
         # Fetch score matrices from computation and scenario characteristics
-        multi_partner_learning = the_scenario.mpl
-        score_matrix_collective_models = multi_partner_learning.score_matrix_collective_models[:, 1:]
-        score_matrix_per_partner = multi_partner_learning.score_matrix_per_partner
+        multi_partner_learning = self.scenario.mpl
+        score_matrix_collective_models = multi_partner_learning.history.history['model']['val_accuracy'][:, 1:]
```
Thomas-Galtier, line 1097: line 1096 triggers an error in line 1109. With minibatch 2 and 10 epoch `score_matrix_collective_models` should be of size 20 but was of size 10.

mplc/multi_partner_learning.py
arthurPignet: Done !

mplc/mpl_utils.py
arthurPignet: Hum really curious
Which sequential exactly ?

mplc/mpl_utils.py
arthurPignet: yes indeed it will be greater for readibility. But I am wondering, isn't it a really long string for a key ?

mplc/mpl_utils.py
arthurPignet: hm it is quite the same. The idea of the `>=`  is to match with the condition for the while loop in fit_epoch.
And the `>=` condition on the while loop is to catch unexpected future uses, as increasing the minibatch_index 2 by 2, and so on

But I don't see what could be the error raised

mplc/mpl_utils.py
arthurPignet: It is made to generate an easy to use pandas dataframe, with all the result.
The previous matrix score, or the new history dict is not really user friendly. I will show an use case of this method in the first tutorial, which need to be update. 

The dataframe could be used like history_df.group_by('Partner').where( ... ) and so on

mplc/mpl_utils.py
bowni: I don't think it is a problem. But to shorten it we could try `mpl_model`?

mplc/mpl_utils.py
bowni: Ok, makes sense

mplc/mpl_utils.py
arthurPignet: ok for me, I go for mpl_model

mplc/mpl_utils.py
Thomas-Galtier: seq-pure

mplc/multi_partner_learning.py
```diff
@@ -60,594 +53,427 @@ def __init__(self,
...
+
+        # Iterate over mini-batches and train
+        for i in range(self.minibatch_count):
+            logger.info(f"(seq-final-agg) Minibatch n°{self.minibatch_index} of epoch n°{self.epoch_index}, "
+                        f"init model with a copy of the global model")
+            self.minibatch_index = i
+            self.fit_minibatch()
+
+        # At the end of each epoch, aggregate the models
+        self.model_weights = self.aggregator.aggregate_model_weights()
```
bowni, line 423: Hmm, small detail that had its importance:

- in seq-with-final-agg, the aggregation is made as I explained last time for having a more reliable end-of-epoch perf eval for early stopping (and final perf test)
- however, initially what we had designed was that the next epoch shouldn't start from the aggregated model. It should start from the final sequential model of the last epoch. So that the learning stays really sequential
- however, thinking about it, I kind of like this intermediate implementation where it starts from the aggregated model. We could keep it like that (although it probably means updating the schema in the documentation)

mplc/multi_partner_learning.py
```diff
@@ -60,594 +53,427 @@ def __init__(self,
...
+            self.minibatch_index = i
+            self.fit_minibatch()
+
+            # At the end of each minibatch, aggregate the models
+            self.model_weights = self.aggregator.aggregate_model_weights()
 
 
 def init_multi_partner_learning_from_scenario(scenario, is_save_data=True):
-    mpl = MultiPartnerLearning(
+    mpl = scenario.multi_partner_learning_approach(
```
bowni, line 471: It feels a bit weird to have this construction method inbetween different MPL approaches

bowni (commented):
A few additional comments.
(I didn't run tests et, will do asap)

README.md
```diff
@@ -76,13 +74,13 @@ Finally, with given scenarios and multi-partner learning approaches, we can addr
...
 
 ##### Build from Source 
 
 If you want to install mplc from the repository, make sure that you got the latest version of pip. 
 
-Then clone the repository, and trigger the installation using pip.
+Then clone the repository, and trigger the installation from local sources.
 
 ```bash
 $ git clone https://github.com/SubstraFoundation/distributed-learning-contributivity.git
```
bowni, line 86: See above/below: there is a duplication of the `### Installation` section

mplc/multi_partner_learning.py
arthurPignet: Yes I agree. 
Unfortunately as Python doesn't allow multi constructors, I personally don't see any other solution that this one. 
Any idea ?

arthurPignet: In fact a mpl object needs a list of partner. The partners are only usable after being instancied in a scenario (instanciation which split the data between the partner). So we will always need a scenario object somehow. But we can want the parameter of the mpl to be different from those set in the scenario object. I suggest to allow this kind of changes via the **kwargs

bowni: What I meant was not to not have it, I remember why we introduced it and I think it makes perfect sense. I meant that it would better to have either before or after all MPL child classes, not in the middle of them

mplc/partner.py
arthurPignet: the function build_model_from_weights is the mpl's one, it needs the model generator provides by the mpl's dataset. 
And the weights are those of the partner... 
Otherwise, I don't really know if it's a proper way to implement that

mplc/multi_partner_learning.py
arthurPignet: aaah ok
By the time I removed this function ... what do you think ?

mplc/mpl_utils.py
```diff
@@ -0,0 +1,177 @@
...
+        plt.title("Model accuracy")
+        plt.ylabel("Accuracy")
+        plt.xlabel("Epoch")
+        plt.legend()
+        plt.ylim([0, 1])
+        plt.savefig(self.save_folder / "graphs/all_partners.png")
+        plt.close()
+
+
+class Aggregator(ABC):
```
bowni, line 123: Good idea!

mplc/multi_partner_learning.py
```diff
@@ -5,804 +5,478 @@
...
-
-        logger.info("Training and evaluation on multiple partners: done.")
-        end = timer()
-        self.learning_computation_time = end - start
+MULTI_PARTNER_LEARNING_APPROACHES = {
+    "fedavg": FederatedAverageLearning,
+    "seq-pure": SequentialLearning,
+    "seq-with-final-agg": SequentialWithFinalAggLearning,
+    "seqavg": SequentialAverageLearning
+}
```
bowni, line 482: Why don't you add `MplLabeLFlip` in there?
(For `SinglePartnerLearning` I suppose it's ok... 😄 )

bowni (approved):

mplc/multi_partner_learning.py
```diff
@@ -5,804 +5,478 @@
...
-                elif self.learning_approach in ['seq-pure', 'seq-with-final-agg']:
-                    self.compute_collaborative_round_sequential(sequentially_trained_model)
+            if self.early_stop():
+                break
+            self.epoch_index += 1
 
-                elif self.learning_approach == 'seqavg':
-                    self.compute_collaborative_round_seqavg()
+        # After last epoch or if early stopping was triggered, evaluate model on the global testset
+        self.history.log_final_model_perf()
```
RomainGoussault, line 166: cf previous comment, we should evaluate the model with self.evaluate() then log the perf.

mplc/mpl_utils.py
```diff
@@ -0,0 +1,177 @@
...
+            self.history[partner_id][key][self.mpl.epoch_index, self.mpl.minibatch_index] = history[key][-1]
+
+        epoch_nb_str = f"Epoch {str(self.mpl.epoch_index).zfill(2)}/{str(self.mpl.epoch_count - 1).zfill(2)}"
+        mb_nb_str = f"Minibatch {str(self.mpl.minibatch_index).zfill(2)}/{str(self.mpl.minibatch_count - 1).zfill(2)}"
+        partner_id_str = f"Partner partner_id #{partner_id} ({partner_index}/{self.mpl.partners_count - 1})"
+        val_acc_str = f"{round(history['val_accuracy'][-1], 2)}"
+
+        logger.debug(f"{epoch_nb_str} > {mb_nb_str} > {partner_id_str} > val_acc: {val_acc_str}")
+
+    def log_model_val_perf(self):
```
RomainGoussault, line 43: same as comment below we should log not evaluate here

mplc/scenario.py
```diff
@@ -17,42 +17,44 @@
 from sklearn.model_selection import train_test_split
 from sklearn.preprocessing import LabelEncoder
 
-from . import contributivity, constants, dataset as dataset_mod
-from . import multi_partner_learning
+from . import contributivity, constants
+from . import dataset as dataset_mod
```
RomainGoussault, line 21: why dataset_mod ?

RomainGoussault (changes requested):
Huge work !

mplc/mpl_utils.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def aggregate_model_weights(self):
+        self.prepare_aggregation_weights()
+        super(ScoresAggregator, self).aggregate_model_weights()
+
+
+# Supported aggregation weights approaches
+AGGREGATORS = {
+    "uniform": UniformAggregator,
+    "data-volume": DatavolumeAggregator,
```
RomainGoussault, line 175: change main config file accordinly

mplc/scenario.py
arthurPignet: shortcut for dataset module, to differ from real dataset.

mplc/mpl_utils.py
```diff
@@ -0,0 +1,177 @@
...
+
+        :type mpl: MultiPartnerLearning
+        """
+
+        self.mpl = mpl
+        self.save_folder = mpl.save_folder
+        self.nb_epochs_done = 0
+        self.score = None  # Final score evaluated on the test dataset at the end of the training
+        self.metrics = ['val_accuracy', 'val_loss', 'loss', 'accuracy']
+        tab = {key: np.nan * np.zeros((mpl.epoch_count, mpl.minibatch_count)) for key in self.metrics}
```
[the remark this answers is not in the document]
arthurPignet: a temporary table

mplc/multi_partner_learning.py
arthurPignet: Ah it is a good question. I thought about it as a contributivity method, not as a mpl one. But it can be interesting to add it yes ! I wonder what append if we compute Shapley with Lflip

mplc/contributivity.py
```diff
@@ -1137,67 +1134,64 @@ def flip_label(self):
...
         elif method_to_compute == "WR_SMC":
             # Contributivity 9: Without replacement Stratified Monte Carlo
-            self.without_replacment_SMC(current_scenario, sv_accuracy=sv_accuracy, alpha=alpha)
+            self.without_replacment_SMC(sv_accuracy=sv_accuracy, alpha=alpha)
         elif method_to_compute == "Federated SBS linear":
             # Contributivity 10: step by step increments with linear importance increase
-            if current_scenario.multi_partner_learning_approach != "fedavg":
+            if self.scenario.multi_partner_learning_approach != multi_partner_learning.FederatedAverageLearning:
                 logger.warning("Step by step linear contributivity method is only suited for federated "
                                "averaging learning approach")
```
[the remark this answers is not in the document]
arthurPignet: Hm I am not sure to understand what you mean. In fact we log this warning only if the mpl chosen is not fedavg.

mplc/mpl_utils.py
arthurPignet: It is not very clear for me what I could change in the main.py file. Because from a scenario's point of view nothing change, and main.py only deals with scenario.

mplc/multi_partner_learning.py
```diff
@@ -110,28 +127,50 @@ def save_final_model(self):
...
-    def build_model_from_weights(self, new_weights):
-        """Generate a new model initialized with weights passed as arguments"""
-        new_model = self.generate_new_model()
-        new_model.set_weights(new_weights)
-        return new_model
+        logger.debug(f"{epoch_nb_str} > {mb_nb_str} > {partner_id_str} > val_acc: {val_acc_str}")
 
-    def init_model(self):
-        new_model = self.generate_new_model()
+    def log_model_val_perf(self):
```
RomainGoussault, line 142: I would cut this method in two:

- `evaluate_val_perf() `which returns a hist object
- `log_perf(hist)`

mplc/multi_partner_learning.py
```diff
@@ -110,28 +127,50 @@ def save_final_model(self):
...
+                              verbose=0,
+                              )
+        self.history.history['mpl_model']['val_loss'][self.epoch_index, self.minibatch_index] = hist[0]
+        self.history.history['mpl_model']['val_accuracy'][self.epoch_index, self.minibatch_index] = hist[1]
+
+        if self.minibatch_index >= self.minibatch_count - 1:
+            logger.info(f"   Model evaluation at the end of the epoch: "
+                        f"{['%.3f' % elem for elem in hist]}")
+
+    def log_final_model_perf(self):
```
RomainGoussault, line 156: Same comment as above

mplc/multi_partner_learning.py
RomainGoussault: Also can we indicate it's local or global validation dataset ?

RomainGoussault (changes requested):

mplc/multi_partner_learning.py
arthurPignet: My idea is that it already exists a evaluate method, the Keras one, so I prefer not duplicate it.

arthurPignet: I can add a comment, if you want. As we use the global method (mpl's one), it is computed on the global set. but maybe it is not so clear

mplc/multi_partner_learning.py
```diff
@@ -5,804 +5,518 @@
...
+                    f"init each partner's models with a copy of the global model")
+
+        for partner in self.partners_list:
+            partner.model_weights = self.model_weights
 
         # Evaluate and store accuracy of mini-batch start model
-        model_to_evaluate = partners_model_list_for_iteration[0]
-        model_evaluation_val_data = self.evaluate_model(model_to_evaluate, self.val_data)
-        self.score_matrix_collective_models[epoch_index, minibatch_index] = model_evaluation_val_data[1]
+        self.history.log_model_val_perf()
```
RomainGoussault: This method does not exist anymore ?

mplc/multi_partner_learning.py
RomainGoussault: Other solution: rename this method to something like: `evaluate_on_val_data(`) ?

RomainGoussault: Yes a comment would be a great for the global/local validation.

arthurPignet: What method ? the log ?

mplc/multi_partner_learning.py
arthurPignet: Yes I fixed it

natct10 (commented):
LGTM but tests are failing on `mnist` & `contrib`

mplc/multi_partner_learning.py
arthurPignet: If you want we can indicate in the method name that we are doing evaluation. But If someone wants to evaluate the model, he will go for the model.evaluate() method, with the data he wants to use. The log methods should never be called by an user, but only on purpose, during a home-build training (like in pvrl). The objective of this log methods is to centralized the history maintaining, and to provide an easy-to-use way of having proper logs and history while designing new contrib or mpl without having to worries about those. It kind of encapsulates the log handling. 

By cutting these methods in two, we would end up with evaluate_on_test_set() and evaluate_on_val_set(), which will be used only once by the logs methods.
Moreover these will be the equivalents of model.evaluate(val_data) and model.evaluate(test_data)

Therefore I think we should kept the method in one piece. 
But we can change their names if the 'log' is strange

natct10:
> LGTM but tests are failing on `mnist` & `contrib`

natct10:
My bad, misclick!

natct10:
The mnist test is failing as the min score is 0.92. Do you want what's happening?
Should we lower the test threshold to 0.92?
```sh
        df = Test_EndToEndTest.get_latest_dataframe("*end_to_end_test*")
        # Extract score
        min_test_score = df["mpl_test_score"].min()
>       assert min_test_score > 0.95
E       assert 0.9215999841690063 > 0.95
tests/end_to_end_tests.py:43: AssertionError
```

mplc/multi_partner_learning.py
RomainGoussault: I agree to keep the method in one piece.
I think we should definitely change the name since the method is not only logging things. What about renaming the method to `evaluate_on_val_data() `? Any idea @natct10 @bowni ?

natct10: > I agree to keep the method in one piece.
> I think we should definitely change the name since the method is not only logging things. What about renaming the method to `evaluate_on_val_data() `? Any idea @natct10 @bowni ?

It is logging:

```python
        if self.minibatch_index >= self.minibatch_count - 1:
            logger.info(f"   Model evaluation at the end of the epoch: "
                        f"{['%.3f' % elem for elem in hist]}")
```

natct10: Ah okay, got it, so let's rename it as suggested

natct10:
I have the same error than travis:
```sh
ERROR    | __main__:<module>:115 - An error has been caught in function '<module>', process 'MainProcess' (231709), thread 'MainThread' (139951218480960):
Traceback (most recent call last):

> File "main.py", line 115, in <module>
    main()
    └ <function main at 0x7f48d1e54e50>

  File "main.py", line 78, in main
    current_scenario.run()
    │                └ <function Scenario.run at 0x7f485d45c040>
    └ <mplc.scenario.Scenario object at 0x7f485cd08940>

  File "/home/zen/dev/distributed-learning-contributivity/mplc/scenario.py", line 849, in run
    self.mpl.fit()
    │    │   └ <function MultiPartnerLearning.fit at 0x7f4866d390d0>
    │    └ <mplc.multi_partner_learning.FederatedAverageLearning object at 0x7f4855ca6e20>
    └ <mplc.scenario.Scenario object at 0x7f485cd08940>

  File "/home/zen/dev/distributed-learning-contributivity/mplc/multi_partner_learning.py", line 207, in fit
    self.save_final_model()
    │    └ <function MultiPartnerLearning.save_final_model at 0x7f4866d36d30>
    └ <mplc.multi_partner_learning.FederatedAverageLearning object at 0x7f4855ca6e20>

  File "/home/zen/dev/distributed-learning-contributivity/mplc/multi_partner_learning.py", line 128, in save_final_model
    model_to_save.save_weights(os.path.join(model_folder, self.dataset_name + '_final_weights.h5'))
    │             │            │  │    │    │             │    └ 'mnist'
    │             │            │  │    │    │             └ <mplc.multi_partner_learning.FederatedAverageLearning object at 0x7f4855ca6e20>
    │             │            │  │    │    └ '/home/zen/dev/distributed-learning-contributivity/experiments/end_to_end_test_mnist_2020-11-03_10h36/scenario_1_repeat_1_202...
    │             │            │  │    └ <function join at 0x7f48ee3875e0>
    │             │            │  └ <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>
    │             │            └ <module 'os' from '/usr/lib/python3.8/os.py'>
    │             └ <function Network.save_weights at 0x7f4866e7fc10>
    └ <keras.engine.sequential.Sequential object at 0x7f48442a43a0>

  File "/home/zen/.local/lib/python3.8/site-packages/keras/engine/saving.py", line 449, in save_wrapper
    save_function(obj, filepath, overwrite, *args, **kwargs)
    │             │    │         │           │       └ {}
    │             │    │         │           └ ()
    │             │    │         └ True
    │             │    └ '/home/zen/dev/distributed-learning-contributivity/experiments/end_to_end_test_mnist_2020-11-03_10h36/scenario_1_repeat_1_202...
    │             └ <keras.engine.sequential.Sequential object at 0x7f48442a43a0>
    └ <function Network.save_weights at 0x7f4866e7fb80>
  File "/home/zen/.local/lib/python3.8/site-packages/keras/engine/network.py", line 1184, in save_weights
    saving.save_weights_to_hdf5_group(f, self.layers)
    │      │                          │  │    └ <property object at 0x7f4866d6af40>
    │      │                          │  └ <keras.engine.sequential.Sequential object at 0x7f48442a43a0>
    │      │                          └ <Closed HDF5 file>
    │      └ <function save_weights_to_hdf5_group at 0x7f4866e7e940>
    └ <module 'keras.engine.saving' from '/home/zen/.local/lib/python3.8/site-packages/keras/engine/saving.py'>
  File "/home/zen/.local/lib/python3.8/site-packages/keras/engine/saving.py", line 760, in save_weights_to_hdf5_group
    param_dset = g.create_dataset(name, val.shape,
                 │ │              │     │   └ <attribute 'shape' of 'numpy.ndarray' objects>
                 │ │              │     └ array([[[[-0.03707604, -0.03573737,  0.13777296,  0.17350416,
                 │ │              │                  0.02721884,  0.07960887,  0.01194842,  0.05688735,
                 │ │              │        ...
                 │ │              └ b'conv2d_83/kernel:0'
                 │ └ <function Group.create_dataset at 0x7f48c1592310>
                 └ <Closed HDF5 group>
  File "/home/zen/.local/lib/python3.8/site-packages/h5py/_hl/group.py", line 143, in create_dataset
    if '/' in name:
              └ b'conv2d_83/kernel:0'

TypeError: a bytes-like object is required, not 'str'

```

natct10:
=> `h5py==2.10.0`

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+h5py==2.10.0
```
natct10, line 1: +1

dev-requirements.txt
```diff
@@ -15,3 +16,5 @@ flake8==3.8.3
 twine==3.2.0
 setuptools==50.3.0
 wheel==0.35.1
+
+
```
natct10, line 20: only one extra line, please

natct10 (approved):
Well done, sir!

RomainGoussault (approved):

PR #256, opened 2020-10-19
End to end titanic (logistic regression) test

RomainGoussault:


natct10 (approved):
Looks good, thank you @RomainGoussault!

tests/end_to_end_tests.py
```diff
@@ -40,6 +40,20 @@ def test_mnist(self):
...
 
+    def test_titanic(self):
+        """
+        Test performance on titanic dataset
+        """
+        # run test
+        subprocess.run(["python", "main.py", "-f", "tests/config_end_to_end_test_titanic.yml"])
+
+        df = Test_EndToEndTest.get_latest_dataframe("*end_to_end_test*")
+
```
arthurPignet, line 51: Is it possible to use the package version of the library ?
Like : 
```python
from mplc.scenario import Scenario

scenario1 = Scenario(2, [0.4,0.6], epoch_count=3, minibatch_count=4)
scenario1.run()
assert  min(scenario1.mpl.test_score) > 0.6 
```

natct10: Oh yes indeed!

arthurPignet (approved):

PR #260, opened 2020-10-27
Fix the shuffling of labels

arthurPignet:
The shuffle of labels for a shuffled partner doesn't work.

### DIFF
--- a/mplc/partner.py
+++ b/mplc/partner.py
@@ -3,7 +3,7 @@
 This enables to parameterize the partners that participate to the simulated federated learning scenario.
 """
 
-from random import shuffle, sample
+from random import sample
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


mplc/partner.py
```diff
@@ -54,19 +54,6 @@ def shuffle_labels(self, proportion_shuffled):
...
-            label = self.y_train[i]
-            # Suffle the label
-            new_label = shuffle(label)
-
-            # Force the label to be different
-            while np.all(new_label == label):
-                new_label = shuffle(label)
-
-            self.y_train[i] = new_label
+        np.random.shuffle(self.y_train[:n])
```
Thomas-Galtier, line 59: This won't work if your partner as only one class.

mplc/partner.py
```diff
@@ -54,19 +54,7 @@ def shuffle_labels(self, proportion_shuffled):
...
-            label = self.y_train[i]
-            # Suffle the label
-            new_label = shuffle(label)
-
-            # Force the label to be different
-            while np.all(new_label == label):
-                new_label = shuffle(label)
-
-            self.y_train[i] = new_label
+        for label in self.y_train[:n]:
```
RomainGoussault, line 59: We want to shuffle only the first n labels ? 
I think it makes more sense to randomly choose n labels among len(y_train) and to shuffle them.

mplc/partner.py
```diff
@@ -54,19 +54,7 @@ def shuffle_labels(self, proportion_shuffled):
             raise ValueError(
                 f"The proportion of labels to corrupted was {proportion_shuffled} but it must be between 0 and 1."
             )
-
         # Select the indices where the label will be shuffled
```
RomainGoussault, line 57: We can delete this comment I guess

RomainGoussault (changes requested):

mplc/partner.py
arthurPignet: I though labels were shuffled during the data splittin between partners, and before each minibatch split. Nevertheless, in advanced split, it is not obvious that the data are shuffled. I changed that

bowni (approved):

natct10 (approved):

RomainGoussault (approved):
👍

PR #275, opened 2020-11-06
Experiment object

### LINKED ISSUE #273: Add an Experiment object to gather scenarios and run them in sequence
@RomainGoussault @arthurPignet what do you think of the following:

- an `Experiment` object, which would essentially contain a list of `Scenario` instances (let's call it `scenarios_list` here)
- a `run()` method which loops over the `scenarios_list` and run them one by one (or in parallel if someone knows how to parallelize computations... see #49 if you have ideas about that)
- as currently done in `main.py` it would add one level in the folders to gather results, corresponding to the `Experiment` instance (e.g. `./experiments/my_experiment/scenario_n/...`)
- additionnally we could have:
  - a `nb_repeats` parameter, as @RomainGoussault you had introduced
  - a `name`
  - a pre-validation of all scenarios, as we have in `main.py`
  - a results file with all scenarios results, as we have in `main.py`
  - a notebook that offers default tables and graphs on the results

It would enable to refactor `main.py` properly, to be much closer to the way the library is supposed to be used in a notebook (for an individual scenario or a serie of scenarios gathered in an `Experiment`).

bowni:
Closes #273

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -22,8 +22,8 @@ scenario_params_list:
      - 'uniform'
    epoch_count: 
      - 38
-   methods:
-     - ["Shapley values", "Independent scores", "TMCS"]
+   contributivity_methods:
+     - [ "Shapley values", "Independent scores", "TMCS" ]
    minibatch_count: 
      - 20
    gradient_updates_per_pass_count:
--- a/main.py
+++ b/main.py
@@ -8,8 +8,8 @@ A script for:
 
 from loguru import logger
 
-from mplc import experiment
 from mplc import utils
+from mplc.experiment import init_experiment_from_config_file
 from mplc.utils import parse_command_line_arguments
 
 DEFAULT_CONFIG_FILE = "./config.yml"
@@ -28,14 +28,11 @@ def main():
     # Initialize GPU configuration
     utils.init_gpu_config()
 
-    # Instantiate an Experiment
-    main_experiment = experiment.Experiment()
-
     # Initialize experiment from configuration file
     config_filepath = args.file if args.file else DEFAULT_CONFIG_FILE
     if not args.file:
         logger.info(f"No config file specified, using default config file: {DEFAULT_CONFIG_FILE}")
-    main_experiment.init_experiment_from_config_file(config_filepath)
+    main_experiment = init_experiment_from_config_file(config_filepath)
 
     # Run the experiment
     main_experiment.run()
--- a/main.py
+++ b/main.py
@@ -99,13 +99,8 @@ def validate_scenario_list(scenario_params_list, experiment_path):
         logger.debug(f"Validation scenario {scenario_id + 1}/{len(scenario_params_list)}")
 
         # TODO: we should not create scenario folder at this point
-        current_scenario = scenario.Scenario(**scenario_params, experiment_path=experiment_path, is_dry_run=True)
-        current_scenario.instantiate_scenario_partners()
-
-        if current_scenario.samples_split_type == 'basic':
-            current_scenario.split_data(is_logging_enabled=False)
-        elif current_scenario.samples_split_type == 'advanced':
-            current_scenario.split_data_advanced(is_logging_enabled=False)
+        current_scenario = scenario.Scenario(**scenario_params, experiment_path=experiment_path)
+        current_scenario.run(is_dry_run=True)
 
     logger.debug("All scenario have been validated")
 
--- a/main.py
+++ b/main.py
@@ -98,9 +98,12 @@ def validate_scenario_list(scenario_params_list, experiment_path):
 
         logger.debug(f"Validation scenario {scenario_id + 1}/{len(scenario_params_list)}")
 
-        # TODO: we should not create scenario folder at this point
-        current_scenario = scenario.Scenario(**scenario_params, experiment_path=experiment_path)
-        current_scenario.run(is_dry_run=True)
+        current_scenario = scenario.Scenario(**scenario_params,
+                                             experiment_path=experiment_path,
+                                             is_dry_run=True,
+                                             )
+        current_scenario.run()
+        current_scenario.is_dry_run = False
 
     logger.debug("All scenario have been validated")
 
--- a/main.py
+++ b/main.py
@@ -6,13 +6,10 @@ A script for:
     - measuring the respective contributions of each partner to the model performance (termed "contributivity")
 """
 
-import os
-
-import matplotlib.pyplot as plt
 from loguru import logger
 
-from mplc import scenario
 from mplc import utils
+from mplc import experiment
 from mplc.utils import parse_command_line_arguments
 
 DEFAULT_CONFIG_FILE = "./config.yml"
@@ -20,92 +17,31 @@ DEFAULT_CONFIG_FILE = "./config.yml"
 
 @logger.catch
 def main():
+
+    # Get arguments
     args = parse_command_line_arguments()
 
-    if args.verbose:
-        debug = True
-    else:
-        debug = False
-
+    # Initialize log level
+    debug = True if args.verbose else False
     utils.init_logger(debug)
-
     logger.debug("Standard output is sent to added handlers.")
 
-    if args.file:
-        logger.info(f"Using provided config file: {args.file}")
-        config = utils.get_config_from_file(args.file)
-    else:
-        logger.info(f"Using default config file: {DEFAULT_CONFIG_FILE}")
-        config = utils.get_config_from_file(DEFAULT_CONFIG_FILE)
-
-    scenario_params_list = utils.get_scenario_params_list(
-        config["scenario_params_list"])
-
-    experiment_path = config["experiment_path"]
-    n_repeats = config["n_repeats"]
-    validate_scenario_list(scenario_params_list, experiment_path)
-
-    for scenario_id, scenario_params in enumerate(scenario_params_list):
-        logger.info(f"Scenario {scenario_id + 1}/{len(scenario_params_list)}: {scenario_params}")
-
-    # Move log files to experiment folder
-    utils.set_log_file(experiment_path)
-
-    # GPU config
+    # Initialize GPU configuration
     utils.init_gpu_config()
 
-    # Close open figures
-    plt.close("all")
+    # Instantiate an Experiment
+    current_experiment = experiment.Experiment()
 
-    # Iterate over repeats of all scenarios experiments
-    for i in range(n_repeats):
+    # Initialize experiment from configuration file
+    config_filepath = args.file if args.file else DEFAULT_CONFIG_FILE
+    if not args.file:
+        logger.info(f"No config file specified, using default config file: {DEFAULT_CONFIG_FILE}")
+    current_experiment.init_experiment_from_config_file(config_filepath)
 
-        logger.info(f"Repeat {i + 1}/{n_repeats}")
-
-        for scenario_id, scenario_params in enumerate(scenario_params_list):
-            logger.info(f"Scenario {scenario_id + 1}/{len(scenario_params_list)}")
-            logger.info("Current params:")
-            logger.info(scenario_params)
-
-            current_scenario = scenario.Scenario(
-                **scenario_params,
-                experiment_path=experiment_path,
-                scenario_id=scenario_id + 1,
-                repeats_count=i + 1
-            )
-
-            current_scenario.run()
-
-            # Write results to CSV file
-            df_results = current_scenario.to_dataframe()
-            df_results["random_state"] = i
-            df_results["scenario_id"] = scenario_id
-
-            with open(experiment_path / "results.csv", "a") as f:
-                df_results.to_csv(f, header=f.tell() == 0, index=False)
-                logger.info(f"Results saved to {os.path.relpath(experiment_path)}/results.csv")
+    # Run the experiment
+    current_experiment.run()
 
     return 0
-
-
-def validate_scenario_list(scenario_params_list, experiment_path):
-    """Instantiate every scenario without running it to check if
-    every scenario is correctly specified. This prevents scenario initialization errors during the experiment"""
-
-    logger.debug("Starting to validate scenarios")
-
-    for scenario_id, scenario_params in enumerate(scenario_params_list):
-
-        logger.debug(f"Validation scenario {scenario_id + 1}/{len(scenario_params_list)}")
-
-        current_scenario = scenario.Scenario(**scenario_params,
-                                             experiment_path=experiment_path,
-                                             is_dry_run=True,
-                                             )
-        current_scenario.run()
-        current_scenario.is_dry_run = False
-
-    logger.debug("All scenario have been validated")
 
 
 if __name__ == "__main__":
--- a/main.py
+++ b/main.py
@@ -8,8 +8,8 @@ A script for:
 
 from loguru import logger
 
-from mplc import utils
 from mplc import experiment
+from mplc import utils
 from mplc.utils import parse_command_line_arguments
 
 DEFAULT_CONFIG_FILE = "./config.yml"
@@ -22,24 +22,23 @@ def main():
     args = parse_command_line_arguments()
 
     # Initialize log level
-    debug = True if args.verbose else False
-    utils.init_logger(debug)
+    utils.init_logger(args.verbose)
     logger.debug("Standard output is sent to added handlers.")
 
     # Initialize GPU configuration
     utils.init_gpu_config()
 
     # Instantiate an Experiment
-    current_experiment = experiment.Experiment()
+    main_experiment = experiment.Experiment()
 
     # Initialize experiment from configuration file
     config_filepath = args.file if args.file else DEFAULT_CONFIG_FILE
     if not args.file:
         logger.info(f"No config file specified, using default config file: {DEFAULT_CONFIG_FILE}")
-    current_experiment.init_experiment_from_config_file(config_filepath)
+    main_experiment.init_experiment_from_config_file(config_filepath)
 
     # Run the experiment
-    current_experiment.run()
+    main_experiment.run()
 
     return 0
 
--- a/mplc/corruption.py
+++ b/mplc/corruption.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 import numpy as np
+from keras.utils import to_categorical
 from loguru import logger
 
 
@@ -156,8 +157,15 @@ class Duplication(Corruption):
             raise Exception('Missing the Partner to duplicate')
         self.generate_matrix()
         idx = self.corrupted_index
+        if self.duplicated_partner.y_train.ndim == 1:
+            self.duplicated_partner.y_train = to_categorical(self.duplicated_partner.y_train.reshape(-1, 1))
+            one_label = True
+        else:
+            one_label = False
         self.partner.y_train[idx] = self.duplicated_partner.y_train[:len(idx)]
         self.partner.x_train[idx] = self.duplicated_partner.x_train[:len(idx)]
+        if one_label:
+            self.duplicated_partner.y_train = np.argmax(self.duplicated_partner.y_train, axis=1).astype('float32')
         logger.debug(f"   Partner #{self.partner.id}: Done.")
 
 
--- a/mplc/doc/documentation.md
+++ b/mplc/doc/documentation.md
@@ -166,7 +166,7 @@ my_scenario = Scenario(partners_count=3,
                        dataset_name='mnist',
                        epoch_count=10,
                        minibatch_count=3,
-                       methods=['Shapley values'])
+                       contributivity_methods=['Shapley values'])
 ```
 
 Once ran, accessing the results is straightforward:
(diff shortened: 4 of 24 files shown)


bowni:
Anyone motivated for creating a default notebook exploring the Experiment results?

mplc/experiment.py
```diff
@@ -0,0 +1,138 @@
...
+
+                # Run the scenario
+                scenario.n_repeat = repeat_idx
+                scenario.run()
+
+                # Save scenario results
+                df_results = scenario.to_dataframe()
+                df_results["repeat_index"] = repeat_idx
+                df_results["scenario_index"] = scenario_idx
+
```
arthurPignet, line 131: We can maybe add somewhere an other .csv per scenario, with the result of the mpl ? from the history object. Otherwise, these will be discarded. 
It can also be a choice to consider these results as too detailed to be stored by an experiment.
I don't know

mplc/experiment.py
```diff
@@ -0,0 +1,138 @@
...
+    def add_scenario(self, scenario_to_add):
+        """Add a scenario to the list of scenarios to be run"""
+
+        if isinstance(scenario_to_add, scenario_module.Scenario):
+            self.scenarios_list.append(scenario_to_add)
+        else:
+            logger.info(f"The scenario {scenario_to_add} you are trying to add is not an instance of"
+                        f"object scenario.Scenario")
+
```
arthurPignet, line 63: Why not raise an error ? I cannot tell why, but an info seems to weak for the message

mplc/experiment.py
```diff
@@ -0,0 +1,138 @@
...
+    ):
+        """
+        :param experiment_name: string, name of the experiment
+        :param scenarios_list: [Scenario], lists instances of the Scenario object
+        :param nb_repeats: int, number of repeats of the experiments (as an experiment includes
+                           a number of non-deterministic phenomena. Example: 5
+        """
+
+        self.name = experiment_name
+        self.scenarios_list = scenarios_list
```
arthurPignet, line 31: What if someone does not give any scenario list ? 
The self.scenario_list will be initialized to `None`, and the `.add_scenario()` will crash with an error like None as no method append, or sort of. It is not really user friendly to debug. I suggest that if no scenario are passed to the Experiment(), the scenario_list should be initialize to []

mplc/scenario.py
```diff
@@ -801,55 +770,66 @@ def to_dataframe(self):
...
         if self.samples_split_type == "basic":
             self.split_data()
         elif self.samples_split_type == "advanced":
             self.split_data_advanced()
-        self.plot_data_distribution()
+
+        if not is_dry_run:
+            self.plot_data_distribution()
         self.compute_batch_sizes()
         self.data_corruption()
```
arthurPignet, line 813: Why don't move these functions to the .__init__ ? 
However the use of the is_dry_run great may be, these methods look like initialization to me ?

arthurPignet (commented):
Nice, this object will help to set up ambitious benchmark

mplc/experiment.py
arthurPignet: outdated

mplc/experiment.py
bowni: I think that is a good idea

mplc/experiment.py
bowni: Yes you are right

mplc/scenario.py
bowni: That is indeed a good point.
We should probably keep the plot out of it, as a separate method available in the SDK for users, but not in the init. Unless it includes saving the plot in the results folder, in which case it makes sense to have it generated (but not displayed by the python IDE / env).

Yes you are right, it doesn't bring any benefit to keep it out. Maybe only the posibility to execute custom ways to do these things. But that won't happen often...

mplc/experiment.py
bowni: @arthurPignet does the `History` object has a method to print it / write it to disk?

arthurPignet: yes, history.save_data() (which is a refacto for the old ml.save_data())
But it saves the dict history.history in binary, I think we should change that to save the dataframe 
The method also save some plots.

mplc/scenario.py
arthurPignet: Yes I am not sure for the handling of the plot.
For the other initialization step, it would be great to have them all in the init. Otherwise, a mpl needs a scenario which have been ran, which is strange. 
For instance, when working with a notebook on a new mpl method, I need a scenario to initialize it, and I need to run it to initialize the partners.

mplc/experiment.py
```diff
@@ -0,0 +1,138 @@
...
+DEFAULT_CONFIG_FILE = "../config.yml"
+
+
+class Experiment:
+    def __init__(
+            self,
+            experiment_name=None,
+            nb_repeats=1,
+            scenarios_list=[]
+    ):
```
arthurPignet, line 25: Maybe I can be great to allow one to pass the path as a parameter ? Maybe via **kwargs ?

bowni: Yes, good idea

mplc/experiment.py
```diff
@@ -20,18 +21,23 @@ def __init__(
...
         """
 
         self.name = experiment_name
-        self.scenarios_list = []
-        self.nb_repeats = nb_repeats
         if experiment_name:
             self.experiment_path = self.define_experiment_path()
+        self.scenarios_list = []
+        for scenario in scenarios_list:
+            self.add_scenario(scenario)
```
bowni, line 38: I am not too fan of that solution. Because now a user could (naively, but realistically) do `my_exp.scenarios_list.append(my_new_scenario)`, and in that case your `save_folder` will not be initialized properly.

mplc/experiment.py
```diff
@@ -115,13 +120,14 @@ def run_experiment(self):
             logger.info(f"(Experiment {self.name}) Now starting repeat {repeat_index_str}")
 
             # Loop over scenarios in scenarios_list
-            for scenario_idx, scenario in enumerate(self.scenarios_list):
+            for scenario_idx, white_scenario in enumerate(self.scenarios_list):
```
bowni: Fun color! I never heard about prefixing a variable with `white_...`, what does it correspond to?

mplc/multi_partner_learning.py
```diff
@@ -210,10 +224,8 @@ def fit(self):
         self.learning_computation_time = end - start
         logger.info(f"Training and evaluation on multiple partners: "
                     f"done. ({np.round(self.learning_computation_time, 3)} seconds)")
-        if self.is_save_data:
-            self.history.save_data()
-            # Save the model's weights
-            self.save_final_model()
+        if self.save_folder is not None:
+            self.save_data()  # Save the model weights and the history data
```
bowni, line 228: I think you can do `if self.save_folder` and omit `is not None`, no?

mplc/scenario.py
```diff
@@ -78,11 +77,10 @@ def __init__(
         :param minibatch_count: int
         :param epoch_count: int
         :param is_early_stopping: boolean. Stop the training if scores on val_set reach a plateau
-        :param methods: A declarative list `[]` of the contributivity measurement methods to be executed.
+        :param contributivity_methods: A declarative list `[]` of the contributivity measurement contributivity_methods to be executed.
         :param is_quick_demo: boolean. Useful for debugging
-        :param experiment_path: path
+        :param save_path: path where to save the scenario output. By default, the scenario is not saved !
```
bowni, line 82: `path where to save the scenario outputs. By default, they are not saved!`

mplc/scenario.py
```diff
@@ -304,13 +304,21 @@ def __init__(
...
-                             f"{uuid.uuid4().hex[:3]}"  # to distinguish identical names
-        self.short_scenario_name = f"{self.partners_count} {self.amounts_per_partner}"
-
-        self.experiment_path = experiment_path
-        self.save_folder = self.experiment_path / self.scenario_name
-
+        self.scenario_name = kwargs.get('scenario_name',
+                                        f"scenario_{self.scenario_id}_repeat_{self.repeat_count}_{now_str}_"
+                                        f"{uuid.uuid4().hex[:3]}")  # to distinguish identical names
+        if ' ' in self.scenario_name:
```
bowni, line 310: Doesn't seem robust - I'd suggest regex in the form of `r'\s'` or something like that

mplc/scenario.py
```diff
@@ -304,13 +304,21 @@ def __init__(
...
+                f'The scenario name "{self.scenario_name}"cannot be written with space character, please use '
+                f'underscore or dash.')
+        self.short_scenario_name = f"{self.partners_count}_{self.amounts_per_partner}"
+
+        if save_path is not None:
+            self.experiment_path = Path(save_path)
+            self.save_folder = self.experiment_path / self.scenario_name
+        else:
+            self.experiment_path = None
+            self.save_folder = None
```
bowni, line 321: Hmmm, so a `scenario` instance gets an `experiment_path` attribute? A bit weird? To me either it is added to an experiment, in which case the experiment `experiment_path` attribute exists, or it is not and in that case it is useless. What did I miss?

mplc/scenario.py
```diff
@@ -321,6 +329,38 @@ def __init__(
...
+    @property
+    def nb_samples_used(self):
+        if len(self.partners_list) == 0:
+            return len(self.dataset.x_train)
+        else:
+            return sum([p.final_nb_samples for p in self.partners_list])
+
+    @property
+    def final_relative_nb_samples(self):
+        return [p.final_nb_samples / self.nb_samples_used for p in self.partners_list]
```
bowni, line 341: 👍

mplc/scenario.py
```diff
@@ -805,22 +837,23 @@ def run(self):
...
         # -----------------------------------------------------
 
-        self.mpl = self.multi_partner_learning_approach(self, is_save_data=True)
+        self.mpl = self._multi_partner_learning_approach(self, custom_name='main_mpl')
         self.mpl.fit()
 
         # ----------------------------------------------------------
-        # Instantiate and run the contributivity measurement methods
+        # Instantiate and run the contributivity measurement contributivity_methods
         # ----------------------------------------------------------
```
bowni, line 853: Lacks the dashes `...---...` to make it pretty 😃

mplc/scenario.py
```diff
@@ -805,22 +837,23 @@ def run(self):
         # -----------------
         # Preliminary steps
         # -----------------
-
-        self.plot_data_distribution()
+        if not self.save_folder is None:
```
bowni: I think `if self.save_folder:` does the trick

mplc/scenario.py
```diff
@@ -478,10 +518,7 @@ def split_data_advanced(self, is_logging_enabled=True):
                 amounts_per_partner[p.id] * len(y_train) * final_resize_factor
             )
             p.final_nb_samples_p_cluster = int(p.final_nb_samples / p.cluster_count)
-        self.nb_samples_used = sum([p.final_nb_samples for p in partners_list])
-        self.final_relative_nb_samples = [
-            p.final_nb_samples / self.nb_samples_used for p in partners_list
-        ]
+
 
         # Partners receive their subsets
```
bowni: 1 extra unnecessary blank line

mplc/experiment.py
```diff
@@ -0,0 +1,143 @@
...
+        """Define the path and create folder for saving results of the experiment"""
+
+        now = datetime.datetime.now()
+        now_str = now.strftime("%Y-%m-%d_%Hh%M")
+
+        full_experiment_name = self.name + "_" + now_str
+        experiment_path = Path.cwd() / constants.EXPERIMENTS_FOLDER_NAME / full_experiment_name
+
+        # Check if experiment folder already exists
+        while experiment_path.exists():
```
natct10, line 51: Just wondering, why not a simple `if` check?

mplc/experiment.py
arthurPignet: I will make the scenarios_list private, even if python has a strange way of considering the private aspect of private attributes

mplc/experiment.py
arthurPignet: changed to blank_scenario. The one which is never ran, and which is copied.

mplc/multi_partner_learning.py
arthurPignet: I doesn't really like this option. Indeed python can cast None to False, but it also casts 0, or [], or '' to False. It will work

mplc/scenario.py
arthurPignet: sorry, I didn't understand

mplc/scenario.py
arthurPignet: see above comment

mplc/scenario.py
bowni: The dashets `------` should match the entire length of the comment line

mplc/scenario.py
```diff
@@ -321,17 +321,15 @@ def __init__(
         self.scenario_name = kwargs.get('scenario_name',
                                         f"scenario_{self.scenario_id}_repeat_{self.repeat_count}_{now_str}_"
                                         f"{uuid.uuid4().hex[:3]}")  # to distinguish identical names
-        if ' ' in self.scenario_name:
+        if r'/s' in self.scenario_name:
```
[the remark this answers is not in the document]
arthurPignet: Indeed. In fact even `r'\s' in ...` doesn't work

config.yml
```diff
@@ -22,8 +22,8 @@ scenario_params_list:
      - 'uniform'
    epoch_count: 
      - 38
-   methods:
-     - ["Shapley values", "Independent scores", "TMCS"]
+   contributivity_methods:
+     - [ "Shapley values", "Independent scores", "TMCS" ]
```
natct10, line 26: Sure about this space?

mplc/doc/documentation.md
```diff
@@ -436,6 +436,58 @@ Example: `methods=["Shapley values", "Independent scores", "TMCS"]`
...
   Example: `is_quick_demo=True`
+  
+## Experiments
+
+At some point of your use of the library you might need to launch several scenarios in a rows. It the same spirit, you might want to repeat one scenario's run, to get rid of the randomness of training, and end with more meaningful results.
+
+The `Experiment` object allows you to set a list of scenarios to run, along with a wanted number of repetition.
+Once instancied, scenarios can be added via the `.add_scenario()`.
+Eventually, all the scenarios can be run, and repeated, by calling the `.run()` method. 
+### Experiment's parameters
```
natct10, line 447:
```suggestion

### Experiment's parameters
```

mplc/doc/documentation.md
```diff
@@ -436,6 +436,58 @@ Example: `methods=["Shapley values", "Independent scores", "TMCS"]`
...
+scenario0 = Scenario(4, [0.25]*4)
+scenario1 = Scenario(10, [0.1]*10)
+scenario2 = Scenario(4, [0.8, 0.1, 0.05, 0.05])
+
+exp = Experiment(experiment_name='my_experiment', nb_repeat=10, scenarios_list=[scenario0, scenario1])
+exp.add_scenario(scenario2)
+
+exp.run()
+````
+
```
natct10, line 490:
```suggestion
```

main.py
```diff
@@ -6,109 +6,43 @@
...
-    if args.verbose:
-        debug = True
-    else:
-        debug = False
+    # Get arguments
+    args = parse_command_line_arguments()
 
+    # Initialize log level
+    debug = True if args.verbose else False
     utils.init_logger(debug)
```
RomainGoussault, line 26: `utils.init_logger(args.verbose) `should work ?

main.py
```diff
@@ -6,109 +6,43 @@
...
 
-    # Close open figures
-    plt.close("all")
-
-    # Iterate over repeats of all scenarios experiments
-    for i in range(n_repeats):
-
-        logger.info(f"Repeat {i + 1}/{n_repeats}")
+    # Instantiate an Experiment
+    current_experiment = experiment.Experiment()
```
RomainGoussault, line 33: `current` prefix is a bit weird as there is only one experiment. Would rename to `my_experiment` or something like that

mplc/corruption.py
```diff
@@ -156,8 +157,15 @@ def apply(self):
             raise Exception('Missing the Partner to duplicate')
         self.generate_matrix()
         idx = self.corrupted_index
+        if self.duplicated_partner.y_train.ndim == 1:
+            self.duplicated_partner.y_train = to_categorical(self.duplicated_partner.y_train.reshape(-1, 1))
+            one_label = True
```
RomainGoussault, line 162: What is the meaning of `one_label` ?

mplc/doc/documentation.md
```diff
@@ -436,6 +436,58 @@ Example: `methods=["Shapley values", "Independent scores", "TMCS"]`
 - `is_quick_demo`: `True` or `False` (default)  
   When set to `True`, the amount of data samples and the number of epochs and mini-batches are significantly reduced, to minimize the duration of the run. This is particularly useful for quick demos or simply for debugging.  
   Example: `is_quick_demo=True`
+  
+## Experiments
+
+At some point of your use of the library you might need to launch several scenarios in a rows. It the same spirit, you might want to repeat one scenario's run, to get rid of the randomness of training, and end with more meaningful results.
```
RomainGoussault, line 442: row without s

mplc/doc/documentation.md
```diff
@@ -436,6 +436,58 @@ Example: `methods=["Shapley values", "Independent scores", "TMCS"]`
...
+
+The `Experiment` object allows you to set a list of scenarios to run, along with a wanted number of repetition.
+Once instancied, scenarios can be added via the `.add_scenario()`.
+Eventually, all the scenarios can be run, and repeated, by calling the `.run()` method. 
+### Experiment's parameters
+Few parameters can be set:
+    - `experiment_name`: String, default `None`. Name of the experiment, will be used for the save path. if `None` the result will not be saved on disk
```
RomainGoussault, line 449: I would not do that. Its the easy way to lose 1 day of computation because you forget to put an `experiment_name`.


I propose:

- if no experiment name is provided, generate one with the current date/time
- Decouple experiment name from the saving --> add a parameter to choose if we want to save the results. default is save=True

RomainGoussault: Also have a mechanism to generate a new experiment name (by adding datetime suffix for instance) if there is already a folder with the same experiment name on disk. I already wrote something like that, I don't know if it's still in the codebase.

RomainGoussault (changes requested):
Go go go experiments :)

mplc/corruption.py
arthurPignet: When your label is binary, labels are one dimensional, and not one-ho-encoded. All the corruption methods work with one-hot-encoded labels, so if there is only one label (for instance titanic, survive), we one-hot-encode it, then store the fact that we change the labels into one_label, then corrupt, then convert the labels back to their original format.

mplc/experiment.py
arthurPignet: Because we loop to add `_bis` to the experiment name until the name does not already exist
However, That will be changed. We now use the current data/time

main.py
arthurPignet: I renamed it `main_experiment`

mplc/doc/documentation.md
bowni: I agree with @RomainGoussault , except on one slight detail: to me default should be `is_save=False`. Saving results to disk is using a resource. In our world, in 2021, using resources should not be a default behaviour, it should only be done when explicitly needed and requested (so I believe 😎 ).

mplc/experiment.py
```diff
@@ -84,11 +87,18 @@ def __init__(
...
-        if experiment_name:
+        now = datetime.datetime.now()
+        now_str = now.strftime("%Y-%m-%d_%Hh%M")
+        self.name = experiment_name + "_" + now_str
+
+        self.is_save = is_save
+        if is_save:
             self.experiment_path = self.define_experiment_path(**kwargs)
+        else:
+            self.result = pd.DataFrame({})
```
bowni: I am not sure it should an `else` here for initializing a `self.result` attribute. Why not @arthurPignet doing that in all cases, even when `is_save=True` and results are (also) saved to disk?

mplc/experiment.py
```diff
@@ -84,11 +87,18 @@ def __init__(
                            a number of non-deterministic phenomena). Example: 5
         :param scenarios_list: list, list of scenarios to be run during the experiment.
                                Scenario can also be added via the .add_scenario() method.
+        :param is_save: boolean. If set to True, the experiment will be save on disk.
```
bowni, line 90: Cool! @arthurPignet don't forget to update the parameters in the `documentation.md` accordingly!

mplc/experiment.py
arthurPignet: In fact I changed that, but now I am asking my self. It is useful to kept into memory a copy of something which we are writing on the disk ?

mplc/corruption.py
RomainGoussault: Ok makes sense. Could you add a comment explaining this ?

arthurPignet: I will in the next PR, switch local/global val/test sets, which will present changes in corruptions, if that's ok for you

mplc/experiment.py
bowni: I think so yes. In memory this is for enabling easy analysis / usage of these results, chaining of experiments, etc. On disk this is for persisting those results when needed (to share them with someone later on, include them in a report, etc.).

arthurPignet (approved):

RomainGoussault (approved):
Looks good to me !

PR #276, opened 2020-11-09
New mpl method: average predictions

RomainGoussault:
This new mpl method simply consist of averaging the predictions of the differents partners. It should work well when data is randomly split, it should NOT work well when data is stratified among partners.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,9 +1,8 @@
-experiment_name: my_example
-n_repeats: 5
+experiment_name: ensembling
+n_repeats: 2
 scenario_params_list:
  - dataset_name:
      - 'mnist'
-     - 'cifar10'
    partners_count:
      - 3
    amounts_per_partner: 
@@ -11,45 +10,15 @@ scenario_params_list:
    samples_split_option: 
      - ['basic', 'random']
      - ['basic', 'stratified']
-     - ['advanced', [[4, 'shared'], [6, 'shared'], [4, 'specific']]]
    multi_partner_learning_approach:
-     - 'fedavg'
-     - 'seq-pure'
-     - 'seq-with-final-agg'
-     - 'seqavg'
+     - 'ensembling'
    aggregation_weighting:
-     - 'data-volume'
      - 'uniform'
    epoch_count: 
-     - 38
-   methods:
-     - ["Shapley values", "Independent scores", "TMCS"]
-   minibatch_count: 
-     - 20
-   gradient_updates_per_pass_count:
-     - 8
-   dataset_proportion:
-     - 1
- - dataset_name:
-     - 'mnist'
-   partners_count:
-     - 2
-   amounts_per_partner: 
-     - [0.5, 0.5]
-   samples_split_option:
-     - ['basic', 'stratified']
-   multi_partner_learning_approach:
-     - 'fedavg'
-   aggregation_weighting: 
-     - 'data_volume' 
-     - 'uniform'
-   epoch_count: 
-     - 50
-   methods:
-     - ["Shapley values", "SMCS", "IS_lin_S", "IS_reg_S"]
+     - 10
    minibatch_count: 
      - 10
    gradient_updates_per_pass_count:
-     - 5
+     - 8
    dataset_proportion:
      - 1
--- a/mplc/multi_partner_learning.py
+++ b/mplc/multi_partner_learning.py
@@ -13,6 +13,7 @@ from keras.backend import clear_session
 from keras.callbacks import EarlyStopping
 from loguru import logger
 from sklearn.preprocessing import normalize
+from sklearn.metrics import accuracy_score, log_loss
 
 from . import constants
 from .mpl_utils import History, Aggregator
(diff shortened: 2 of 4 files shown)


config.yml
```diff
@@ -1,55 +1,24 @@
-experiment_name: my_example
```
arthurPignet, line 1: Why did you add changes to this file ? I thought it should stay unchanged

mplc/multi_partner_learning.py
```diff
@@ -139,13 +140,14 @@ def log_partner_perf(self, partner_id, partner_index, history):
...
         model = self.build_model()
-        hist = model.evaluate(self.val_data[0],
-                              self.val_data[1],
-                              batch_size=constants.DEFAULT_BATCH_SIZE,
-                              verbose=0,
-                              )
+        hist = model.evaluate(x, y, batch_size=constants.DEFAULT_BATCH_SIZE, verbose=0)
+
+        return hist
+
```
arthurPignet, line 148: I disagree with this changes. For most of the mpl method, it is just a copy of the keras' one. 
Furthermore, you are not evaluating a model, but a mpl, which I found a little bit strange. 
Finally, it allows you to create a strange mpl method, which do not return any usable model.

The model saved by your method is a nn, which hasn't been fitten and which is far away from the objective of your method. Same for the return of the build_model() method. 

The change of the .evaluate() should be done at the model level I think
For instance you could create a class EnsembleModel

```python
class AveragedModel( ? ):
   def __init__(list_of_partner_model): 
         self.list_of_partner_model = list_of_partner_model
  
   def fit(x, y)
         pass #or maybe train all the partners' models with the new data 
  
   def predict(x):
     predictions_list = []

            model = partner.build_model()
            predictions = model.predict(x)
            predictions_list.append(predictions)

          return np.mean(predictions_list, axis=0)  # by the way, by curiosity can it be done with keras or sklearn ? 
def evaluate(x, y):

# and so on , with the save function, etc
```

mplc/multi_partner_learning.py
```diff
@@ -155,14 +157,9 @@ def eval_and_log_model_val_perf(self):
...
-    def eval_and_log_final_model__test_perf(self):
+    def eval_and_log_final_model_test_perf(self):
         logger.info("### Evaluating model on test data:")
-        model = self.build_model()
-        hist = model.evaluate(self.test_data[0],
-                              self.test_data[1],
-                              batch_size=constants.DEFAULT_BATCH_SIZE,
-                              verbose=0,
-                              )
+        hist = self.evaluate_model(self.test_data[0], self.test_data[1])
```
arthurPignet, line 162: same

mplc/multi_partner_learning.py
```diff
@@ -155,14 +157,9 @@ def eval_and_log_model_val_perf(self):
                         f"{epoch_nb_str}: "
                         f"{['%.3f' % elem for elem in hist]}")
 
-    def eval_and_log_final_model__test_perf(self):
+    def eval_and_log_final_model_test_perf(self):
```
arthurPignet, line 160: see above comment

mplc/multi_partner_learning.py
```diff
@@ -331,6 +328,81 @@ def fit_minibatch(self):
         logger.debug("End of fedavg collaborative round.")
 
 
+class EnsembleResults(MultiPartnerLearning):
+    def __init__(self, scenario, **kwargs):
+        # First, if only one partner, fall back to dedicated single partner function
+        super(EnsembleResults, self).__init__(scenario, **kwargs)
```
arthurPignet, line 334: what about the return of build_model, and the model saved ?

mplc/multi_partner_learning.py
```diff
@@ -331,6 +328,81 @@ def fit_minibatch(self):
...
+    def evaluate_model(self, x, y_true):
+
+        predictions_list = []
+        for partner in self.partners_list:
+
+            model = partner.build_model()
+            predictions = model.predict(x)
+            predictions_list.append(predictions)
+
+        y_pred = np.mean(predictions_list, axis=0)
```
arthurPignet, line 347: this method should be the one of a model object

mplc/multi_partner_learning.py
```diff
@@ -331,6 +328,81 @@ def fit_minibatch(self):
...
+            # Reference the partner's model
+            partner_model = partner.build_model()
+
+            # Train on partner local data set
+            history = partner_model.fit(partner.minibatched_x_train[self.minibatch_index],
+                                        partner.minibatched_y_train[self.minibatch_index],
+                                        batch_size=partner.batch_size,
+                                        verbose=0,
+                                        validation_data=self.val_data)
+
```
arthurPignet, line 396: Since the training of the partners can be done completely separately, I believe it would led to a major speed up to do the whole training at once, in one keras' .fit(), in the way of a single partner training

mplc/partner.py
```diff
@@ -103,4 +103,5 @@ def split_minibatches(self):
         self.minibatched_y_train = np.split(y_train, (split_indices[:-1] * len(y_train)).astype(int))
 
     def build_model(self):
+        # TODO Why don't we have the model as a parameter of partnerMPL object ?
```
arthurPignet, line 106: Because for the keras models we store only the weights and discard the tensorflow graphe to save memory. So we need to rebuild it. 
It is the role of the clear_session() method

tests/end_to_end_tests.py
```diff
@@ -51,6 +51,17 @@ def test_titanic(self):
...
+        """
+        Test performance on mnist dataset with the ensemble method
+        """
+
+        ensemble_scenario = Scenario(2, [0.4, 0.6], epoch_count=3, minibatch_count=1, dataset_name='mnist',
+                                     multi_partner_learning_approach="ensembling")
+        ensemble_scenario.run()
+
+        assert np.min(ensemble_scenario.mpl.history.score) > 0.9
+
```
arthurPignet, line 64: why end_to_end_test and not unit test ?

arthurPignet (changes requested):
Really interesting approach ! I think it could be interesting to use the aggregator weight to do the ensembling, or to see the influence of those weight to the final score, maybe it could lead to a contributivity method ? 
I just wonder about the safety of this method. Is it possible to combine it with differencial privacy ? I don't think that secure aggregation is possible. what do you think ?

mplc/multi_partner_learning.py
arthurPignet: maybe with a [AvgPooling](https://keras.io/api/layers/pooling_layers/average_pooling1d/) layer, or [ensemble method](https://scikit-learn.org/stable/modules/ensemble.html)

It is not really important

PR #279, opened 2020-11-16
Fix early stopping

Thomas-Galtier:
Fix early stopping


bowni (approved):
Seems all good to me, nice catch @Thomas-Galtier ! @arthurPignet can you also review and make sure of the coherence with the mpl refactoring you did?

arthurPignet (approved):

PR #281, opened 2020-11-23
LFlip refacto in S-model

Thomas-Galtier:
S-model is a method like LFlip but it is more robust to random noise

### DIFF
--- a/mplc/scenario.py
+++ b/mplc/scenario.py
@@ -131,7 +131,7 @@ class Scenario:
                          "scenario_name",
                          "repeat_count"]
 
-        unrecognised_parameters = [x for x in kwargs.keys() if x not in params_known]
+        unrecognised_parameters = [x for x in kwargs.keys() if (x not in params_known and not x.startswith('mpl_'))]
         if len(unrecognised_parameters) > 0:
             for x in unrecognised_parameters:
                 logger.debug(f"Unrecognised parameter: {x}")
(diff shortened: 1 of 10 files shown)


mplc/scenario.py
```diff
@@ -349,6 +349,14 @@ def __init__(
         if not is_dry_run:
             self.save_folder.mkdir(parents=True, exist_ok=True)
 
+        # ------------------------------------------------------------------
+        # Select in the kwargs the parameters to be transferred to sub object
+        # ------------------------------------------------------------------
+        self.mpl_kwargs = {}
+        for key, value in kwargs.items():
+            if key.startswith('mpl_'):
+                self.mpl_kwargs[key.replace('mpl_', '')] = value
```
bowni, line 358: Are you sure renaming the arguments (by removing `mpl_`) is needed and bring more advantages than risks of future errors?

bowni (approved):

bowni:
Good for me, apart from a secondary question I raised ☝️ .
Also, just saw that it requires some conflicts to be fixed!

mplc/scenario.py
arthurPignet: the name is really important as we will use this dict as kwargs, with **, and I think that the keyword arguments of mpl should not be all prefixed by mpl_. Moreover I don't think there are risks, because all these renamed parameters are in a specific kwargs. To be honest I am not really a fan of this solution, but I did not find a better one.

arthurPignet (approved):

natct10 (approved):
Looks nice, thank you!

mplc/scenario.py
bowni: OK. So if I understand it correctly: when initializing a scenario, if a user wants to pass kwargs to the mpl that will be instantiated by the scenario, she should prefix them by `mpl_`. Right?
I could work with that. But it implies a very clear documentation for users right?

PR #283, opened 2020-11-27
Change Cifar10's model optimizer

arthurPignet:
The RMSprop optimizer can be changed to adam with a different learning rate. It greatly improves the training performance, as shown below. 

image
image
image

However, even if I tried a few learning rates and decay values, these are propably not the perfect hyperparameters, do not hesitate to suggest another setup. 

### DIFF
--- a/mplc/dataset.py
+++ b/mplc/dataset.py
@@ -21,7 +21,7 @@ from keras.layers import Dense, Dropout
 from keras.layers import Embedding, Conv1D, MaxPooling1D, Flatten
 from keras.losses import categorical_crossentropy
 from keras.models import Sequential
-from keras.optimizers import RMSprop
+from keras.optimizers import Adam
 from keras.preprocessing import sequence
 from keras.utils import to_categorical
 from librosa import load as wav_load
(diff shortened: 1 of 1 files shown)


mplc/dataset.py
```diff
@@ -189,8 +189,8 @@ def generate_new_model(self):
         model.add(Dense(self.num_classes))
         model.add(Activation('softmax'))
 
-        # initiate RMSprop optimizer
-        opt = RMSprop(learning_rate=0.0001, decay=1e-6)
+        # initiate Adam optimizer
+        opt = Adam(learning_rate=0.001, decay=1e-5)
```
RomainGoussault, line 193: https://keras.io/api/optimizers/adam/
what is decay ? it is the beta_1 in the doc ?

mplc/dataset.py
```diff
@@ -189,8 +189,8 @@ def generate_new_model(self):
...
 
         # Let's train the model using RMSprop
```
RomainGoussault, line 195: to remove

RomainGoussault (approved):

Thomas-Galtier (approved):

mplc/dataset.py
arthurPignet: 👍

mplc/dataset.py
arthurPignet: No, beta_1 and beta_2 are for the momentum. 
As far as I understood the doc, the decay is for the learning rate.

Issue #287, opened 2020-12-10
Need a parameter to choose if the val/test sets would be local or global


arthurPignet:
For CIFAR10 and MNIST, we define local and global val/test datasets. Thus we only train on 60% of the dataset, and the local val/test set are never used. 

We need a parameter to choose if these test/val sets would be local or global. 

This parameter should also be used in the log_perf functions, to select the type of dataset on which the scores will be computed. 

For instance a dataset would provide train, test, and val sets. Then when splitting the train dataset between partners, the val and test sets would also be split between partners if the test_set/val_set is 'local'. 

Maybe with two parameters ? like:
validation_set: str, 'local', 'global', 'both' (not a fan of 'both, any suggestion ?)
test_set: str,  'local', 'global', 'both' (not a fan of 'both, any suggestion ?)

If validation_set is 'local', the val_acc, val_loss per partner will be computed locally. And the global_model val perf would be the average of the local perfs. 
For the early-stopping, we could use this averaged score. However it might raises safety issues. 

if test_set is 'local' the final test score is computed as the average of partner's local test score. 

With global, we should keep the current implementation in my opinion. 

I suggest that the local val and test sets are corrupted, if corruption is specified of course. 

What do you think @bowni , @RomainGoussault  ?

PR #288, opened 2020-12-10
Switch local or global val/test set

arthurPignet:
add an option on how compute val score and test score. If set to local, the global test/val set will be split between partners exactly like the train set. it will also be corrupted as the train set is. 

Then the val score and test score for global model will be the average of the scores computed on each partners set. 

 closing #287

### DIFF
--- a/mplc/corruption.py
+++ b/mplc/corruption.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+"""
+This enables to corrupt partner's datasets.
+"""
+
 from abc import ABC, abstractmethod
 
 import numpy as np
--- a/mplc/corruption.py
+++ b/mplc/corruption.py
@@ -241,7 +241,7 @@ class Duplication(Corruption):
         return self._corrupted_test_idx
 
     def set_duplicated_partner(self, partner_list):
-        if not self.duplicated_partner_id:
+        if self.duplicated_partner_id is None:  # We cannot use the `if not self.dupl... ` because if not 0 return True
             raise Exception('Missing the Partner-to-duplicate\'s id.')
         self.duplicated_partner = [partner for partner in partner_list if partner.id == self.duplicated_partner_id][0]
 
--- a/mplc/corruption.py
+++ b/mplc/corruption.py
@@ -16,7 +16,9 @@ class Corruption(ABC):
         self.partner = None
         self.matrix = None
 
-        self._corrupted_idx = None
+        self._corrupted_train_idx = None
+        self._corrupted_val_idx = None
+        self._corrupted_test_idx = None
 
         if partner:
             self.set_partner(partner)
@@ -25,11 +27,25 @@ class Corruption(ABC):
         self.partner = partner
 
     @property
-    def corrupted_index(self):
-        if self._corrupted_idx is None:
+    def corrupted_train_index(self):
+        if self._corrupted_train_idx is None:
             n = int(len(self.partner.y_train) * self.proportion)
-            self._corrupted_idx = np.random.choice(len(self.partner.y_train), size=n, replace=False)
-        return self._corrupted_idx
+            self._corrupted_train_idx = np.random.choice(len(self.partner.y_train), size=n, replace=False)
+        return self._corrupted_train_idx
+
+    @property
+    def corrupted_val_index(self):
+        if self._corrupted_val_idx is None:
+            n = int(len(self.partner.y_val) * self.proportion)
+            self._corrupted_val_idx = np.random.choice(len(self.partner.y_val), size=n, replace=False)
+        return self._corrupted_val_idx
+
+    @property
+    def corrupted_test_index(self):
+        if self._corrupted_test_idx is None:
+            n = int(len(self.partner.y_test) * self.proportion)
+            self._corrupted_test_idx = np.random.choice(len(self.partner.y_test), size=n, replace=False)
+        return self._corrupted_test_idx
 
     @abstractmethod
     def apply(self):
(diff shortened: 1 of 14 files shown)


mplc/scenario.py
```diff
@@ -180,6 +183,15 @@ def __init__(
                 f"Computation use the full dataset for scenario #{scenario_id}"
             )
 
+        if test_set in ['local', 'global', 'both']:
+            self.test_set = test_set
+        else:
+            raise ValueError(f'Test set can be \'local\', \'global\' or \'both\', not {test_set}')
+        if test_set in ['local', 'global', 'both']:
```
bowni, line 190: Here it should be `val_set` which should be tested against the list

mplc/scenario.py
```diff
@@ -550,19 +562,196 @@ def split_data_advanced(self, is_logging_enabled=True):
...
+
+            y_val = LabelEncoder().fit_transform([str(y) for y in self.dataset.y_val])
+
+            # Stratify the dataset into clusters per labels
+            x_val_for_cluster, y_val_for_cluster, nb_samples_per_cluster_val = {}, {}, {}
+            for label in labels:
+                idx_in_full_valset = np.where(y_val == label)
+                x_val_for_cluster[label] = self.dataset.x_val[idx_in_full_valset]
+                y_val_for_cluster[label] = self.dataset.y_val[idx_in_full_valset]
+                nb_samples_per_cluster_val[label] = len(y_val_for_cluster[label])
```
bowni, line 583: This leads to a lot lot of code duplication. Couldn't we factorize `split_data_advanced` in a function which we would apply to the datasets? Or sequence it the other way around, first the split among partners, then the val and test split?
Not sure what would be best, but we should try to avoid duplicating too much code.

bowni (changes requested):

tests/unit_tests.py
```diff
@@ -99,14 +100,15 @@ def create_Partner(create_all_datasets):
...
-                                Cifar10, ["advanced", [[4, "specific"], [6, "shared"], [4, "shared"]]],
-                                ['not-corrupted'] * 3)),
+                params=((Mnist, "random", ['not-corrupted'] * 3),
+                        (Mnist, "random", ['permutation', Redundancy(0.2), Duplication(duplicated_partner_id=0)]),
+                        (Mnist,
+                         AdvancedSplitter([0.3, 0.5, 0.2], [[4, "specific"], [6, "shared"], [4, "shared"]]),
+                         ['not-corrupted'] * 3),
+                        (Cifar10, "random", ['not-corrupted'] * 3),
+                        (Cifar10,
+                         AdvancedSplitter([0.3, 0.5, 0.2], [[4, "specific"], [6, "shared"], [4, "shared"]]),
```
natct10, line 110: So only `AdvancedSplitter` is tested here?

arthurPignet: No, random is tested too

arthurPignet: But indeed we should test stratified too

mplc/corruption.py
```diff
@@ -82,15 +110,27 @@ def generate_matrix(self):
...
+        idx_val = self.corrupted_val_index
+        for i in idx_val:
+            temp = np.zeros((self.partner.num_labels,))
+            temp[np.random.choice(self.partner.num_labels, p=self.matrix[np.argmax(self.partner.y_val[i])])] = 1
+            self.partner.y_val[i] = temp
+        logger.debug(f"   Partner #{self.partner.id}: Done.")
+        idx_test = self.corrupted_test_index
+        for i in idx_test:
+            temp = np.zeros((self.partner.num_labels,))
+            temp[np.random.choice(self.partner.num_labels, p=self.matrix[np.argmax(self.partner.y_test[i])])] = 1
```
natct10, line 131: Need coffee cc @RomainGoussault

mplc/doc/documentation.md
```diff
@@ -232,19 +232,21 @@ There are 2 ways to select a dataset. You can either choose a pre-implemented da
...
+    - `'AdvancedSplitter'`: in certain cases it might be interesting to split the dataset among partners in a more elaborate way. For that we consider the data samples from the initial dataset as split in clusters per data labels. The advanced split is configured by indicating, for each partner in sequence, the following 2 elements: `[[nb of clusters (int), 'shared' or 'specific']]`. Practically, you can either instantiate your `AdvancedSplitter` object, and pass this list `[[nb of clusters (int), 'shared' or 'specific']]` to the keyword argument `description`, or use the string identifier and pass the list `[[nb of clusters (int), 'shared' or 'specific']]` to the scenario via the keyword argument `samples_split_configuration`.
+        String identifier:`'advanced'`.
+        Configuration:  
+        - `nb of clusters (int)`: the given partner will receive data samples from that many different clusters (clusters of data samples per labels/classes)
+        - `'shared'` or `'specific'`:
+         - `'shared'`: all partners with option `'shared'` receive data samples picked from clusters they all share data samples from
```
natct10, line 246: one extra space needed

tests/unit_tests.py
```diff
@@ -324,6 +270,157 @@ def test_generate_new_model(self, create_all_datasets):
...
+    def test_random_splitter_global(self, create_all_datasets):
+        splitter = RandomSplitter([0.1, 0.2, 0.3, 0.4])
+        dataset = create_all_datasets
+        partners_list = [Partner(i) for i in range(len(splitter.amounts_per_partner))]
+        splitter.split(partners_list, dataset)
+        for p in partners_list:
+            assert len(p.y_val) == 0, "validation set is not empty in spite of the val_set == 'global'"
+            assert len(p.y_test) == 0, "test set is not empty in spite of the val_set == 'global'"
+            assert len(p.x_train) == len(p.y_train), 'labels and samples numbers mismatches'
+            assert (p.final_nb_samples / len(dataset.y_train) - splitter.amounts_per_partner[p.id]) \
```
natct10, line 289: Top!

mplc/scenario.py
arthurPignet: Done with splitter object !

bowni (approved):

natct10 (approved):
Nice!

PR #295, opened 2020-12-15
normalize contributivity scores in 3 SBS methods

jeromechambost:
Constraint contributivity scores for step by step methods between 0 and 1


RomainGoussault (approved):
It works well on my machine :)

natct10 (approved):
Not tested but looks good to me!

PR #300, opened 2020-12-21
Test for all contrib methods

RomainGoussault:


tests/contrib_end_to_end_test.py
```diff
@@ -55,3 +56,35 @@ def test_mnist_contrib(self):
...
             assert small_dataset_score.values < big_dataset_score.values
+
+    def test_all_contrib_methods(self):
+        """
+        Test all available contributivity methods on mnist
+        """
+
+        all_methods = constants.CONTRIBUTIVITY_METHODS.copy()
+        all_methods.remove('AIS_Kriging_S')  # This one fails
+
```
arthurPignet, line 67: Why not removing IS reg S ?

arthurPignet (commented):
Really nice ! How does it take to run ?

tests/contrib_end_to_end_test.py
RomainGoussault: Why we should remove them ?

RomainGoussault:
> Really nice ! How long does it take to run ?

It's 10 min now for all the end to end tests

tests/contrib_end_to_end_test.py
arthurPignet: You specified another test with it, and commented that this method would not start with less than 4 partners

arthurPignet (approved):

PR #306, opened 2020-12-23
Average predictions come back

RomainGoussault:


mplc/multi_partner_learning.py
```diff
@@ -603,12 +603,64 @@ def fit_minibatch(self):
...
+
+        if self.partners_count == 1:
+            raise ValueError('Only one partner is provided. Please use the dedicated SinglePartnerLearning class')
+
+        partner_model_list = [self.dataset.generate_new_model() for _ in range(self.partners_count)]
+        self.model = EnsemblePredictionsModel(partner_model_list)
+        # Issue here is that we have 2 models, the Ensemble model and
+        # the model for every partner. self.model could either be the EnsembleModel
+        # or the partner model but in both cases it's not super straightforward.
+
```
RomainGoussault, line 624: @arthurPignet help :)

arthurPignet: I think that the mpl.model should be the Ensemble model, whihc hold as attribute a list of submodel (one per partner)

Therefore in the .fit method, you could do somethink like :

```python
for p in self. partner_list:
      self.model.partners_models[p.id].fit(p.x_train, p.y_train)

```

mplc/multi_partner_learning.py
```diff
@@ -603,12 +605,65 @@ def fit_minibatch(self):
...
+    Ensemble (average) prediction of several input models
+    This approach can only be used with the EnsemblePredictionsModel
+    """
+
+    def __init__(self, scenario, **kwargs):
+        # First, if only one partner, fall back to dedicated single partner function
+        super(EnsemblePredictions, self).__init__(scenario, **kwargs)
+
+        if self.partners_count == 1:
+            raise ValueError('Only one partner is provided. Please use the dedicated SinglePartnerLearning class')
```
bowni, line 619: Maybe move the comment down just before `if self.partners_count == 1:`

mplc/multi_partner_learning.py
```diff
@@ -603,12 +605,65 @@ def fit_minibatch(self):
...
+
+        partner_model_list = [self.dataset.generate_new_model() for _ in range(self.partners_count)]
+        self.model = EnsemblePredictionsModel(partner_model_list)
+
+        for partner in self.partners_list:
+            partner.model_weights = self.model_weights
+        # the partners should all start with a different set of weights
```
bowni, line 626: Hmmm, how does this comment combines with the code around it?

mplc/multi_partner_learning.py
```diff
@@ -603,12 +605,65 @@ def fit_minibatch(self):
...
+    def fit_epoch(self):
+        # Clear Keras' old models
+        clear_session()
 
+        self.eval_and_log_model_val_perf()
+
+        for partner_index, partner in enumerate(self.partners_list):
+
+            partner_model = partner.build_model()  # issue here is it tries to build
+            # the partner model with the Ensembling model
```
bowni, line 642: Is the comment still valid?

bowni, line 642: I don't understand one thing: at the beginning of an epoch here, we are supposed to init the local models with the resulting joint model from the previous epoch, right? How is it achieved here?

bowni (commented):
@RomainGoussault a few questions!

mplc/multi_partner_learning.py
RomainGoussault: Ah yes a TODO that was not done! I should modify that.

mplc/multi_partner_learning.py
RomainGoussault: No, we don't init the local models with the resulting joint model from the previous epoch. Every model training is done independantly. 
You only average the predictions at the very end of the training.

bowni: Oh OK 🤦‍♂️ ! Then follow-up questions: how come you don't pass the `fit()` any parameter for the nb of epochs to train on? Do you rebuild a model for each epoch - which wouldn't make much sense for local trainings right?

RomainGoussault: Yes but this is inside the fit_epoch() function so it should be for one epoch only.
Another solution would be to rewrite the fit() method of MultipartnerLearning for this subclass.

arthurPignet (approved):

mplc/multi_partner_learning/basic_mpl.py
```diff
@@ -660,3 +662,55 @@ def fit_minibatch(self):
...
+
+        # First, if only one partner, fall back to dedicated single partner function
+        if self.partners_count == 1:
+            raise ValueError('Only one partner is provided. Please use the dedicated SinglePartnerLearning class')
+
+        partner_model_list = [self.dataset.generate_new_model() for _ in range(self.partners_count)]
+        self.model = EnsemblePredictionsModel(partner_model_list)
+
+        for partner in self.partners_list:
+            partner.model_weights = self.model_weights
```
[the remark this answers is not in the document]
arthurPignet: The method model_weights call the method .get_weights (for Keras model), which create a list from the tf.tensor
So I don't think it's needed for keras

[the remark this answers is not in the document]
RomainGoussault: I made a deepcopy.
Without the deepcopy the arrays had the same memory address.

bowni (approved):

bowni (changes requested):
What about the documentation? 😃

mplc/doc/documentation.md
```diff
@@ -307,6 +307,8 @@ There are several parameters influencing how the collaborative and distributed l
   - `'fedgrads'`: stands for gradient averaging, quite similar to federated averaging, but the partner-loss's gradients are averaged before the optimization step, instead of averaged the model's weights after the optimization step.
   Warning : This method needs a Keras model to work with. The `gradient_pass_per_update` is set to 1 in the current implementation. 
  
+  - `'ensemble'`: stands for ensembling strategy and is one of the simplest way to train a machine learning model collaboratively. Every parter trains its model independently and the final predictions are the average of every parter's predictions.
```
bowni, line 310:
```suggestion
  - `'ensemble'`: stands for ensembling strategy and is one of the simplest way to train a machine learning model collaboratively. Every partner trains its model independently and the final predictions are the average of every partner's predictions.
```

bowni (approved):
Well done @RomainGoussault 👍

PR #308, opened 2021-01-02
Nightly build

RomainGoussault:


.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,51 @@
+name: nightly-build
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 0 * * *' # run every night ay midnight
```
natct10, line 6: This should work, do you have any other idea why it was not triggered?

bowni:
Now obsolete

PR #314, opened 2021-01-07
Rollback to Python 3.6 & Add 3.6 to gh actions

natct10:
For colab compatibility

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest # We could run on other OS but we would have to change the command to install libsndfile
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.8]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


arthurPignet (approved):
I think this release note will not be too long to write

.github/workflows/main_ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest # We could run on other OS but we would have to change the command to install libsndfile
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.6, 3.8]
```
RomainGoussault, line 11: no 3.7 ?

arthurPignet: We already test the 3.6 and the 3.8, I don't think it's mandatory.
This way the idea is to test min and max supported python's version

natct10: Yes, that was the idea. Are you okay with that @RomainGoussault?

RomainGoussault: Well it's free, so I don't see why we should not test 3.7 :)

arthurPignet: It's a plus 50 percent on the total computation cost so it's not really free.

bowni (approved):

.github/workflows/main_ci.yml
RomainGoussault: By free I meant that all the builds are done in parallel so it does not have any impact on the build time.

natct10: Ok added

PR #318, opened 2021-01-12
Homogenize aggregation and aggregation_weighting

jeromechambost:
Fix aggregation method for experiments, that stayed default "data-volume"

### DIFF
--- a/mplc/contributivity.py
+++ b/mplc/contributivity.py
@@ -977,7 +977,7 @@ class Contributivity:
             logger.info(f"Partners selected for the next epoch: {[p.id for p in mpl.partners_list]}")
 
             # apply one epoch with the selected partner to the previous model/ do the action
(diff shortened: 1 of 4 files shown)


natct10 (approved):
Neat, thank you @jeromechambost!

bowni (approved):
LGTM - @arthurPignet could you please review, and merge if everything looks good to you?

arthurPignet (commented):
I think that the difference between _aggregation_weighting and aggregation_weighting is not easy to understand in a first look.
I would suggest to merge those, as every Aggregator holds an attribute .name. (I think that scenario.aggregation_weighting == scenario._aggregation_weighting.name.

Then we can copy the operation done by corruption, or mpl for the initialization. the parameter aggregation_weighting could be either an aggregator instance, or a string identifier

arthurPignet:
I merged it because I am going to work on aggregators in the PR315

PR #320, opened 2021-01-12
Create a fully flexible splitter

bowni:
The fully flexible splitter enables the definition of a very precise split of a dataset among partners. One can indicate for each partner the % of data samples of each class that the partner should receive.


mplc/scenario.py
```diff
@@ -174,6 +174,11 @@ def __init__(
         # (% of samples of the dataset for each partner, ...
         # ... has to sum to 1, and number of items has to equal partners_count)
         self.amounts_per_partner = amounts_per_partner
+        if np.sum(self.amounts_per_partner) != 1:
+            raise ValueError("The sum of the amount per partners you provided isn't equal to 1")
+        if len(self.amounts_per_partner) != self.partners_count:
+            raise AttributeError(f"The amounts_per_partner list should have a size ({len(self.amounts_per_partner)}) "
+                                 f"equals to partners_count ({self.partners_count})")
```
arthurPignet, line 181: I don't see why you want to move those from the splitter to the scenario ?
I think it might create issues if someone is using a splitter outside a scenario

mplc/splitter.py
```diff
@@ -89,6 +82,69 @@ def copy(self):
...
+                    "split currently discards the amounts_per_partner (if provided) and infers amounts of samples "
+                    "per partner from the samples_split_configuration provided.")
+
+        # First we re-assemble the split configuration per cluster, and check its coherence
+        self.samples_split_grouped_by_cluster = list(zip(*configuration))
+        for idx, cluster in enumerate(self.samples_split_grouped_by_cluster):
+            if np.sum(cluster) > 1:
+                raise ValueError(f"Amounts of samples of class {idx} split among partners exceed 100%, "
+                                 f"the dataset split cannot be performed.")
+
```
arthurPignet, line 100: You could probably vectorize this, as the samples_split_grouped_by_cluster is a list of list (ie np.array)

mplc/splitter.py
```diff
@@ -89,6 +82,69 @@ def copy(self):
...
+        # Init of the superclass to inherit its methods
+        super().__init__(amounts_per_partner, **kwargs)
+
+    def __str__(self):
+        return self.name + str(list(zip(self.samples_split_grouped_by_cluster)))
```
arthurPignet, line 105: This could be re-write with f-string. Moreover, isn't it long ? lists of lists are not really easy to read.

mplc/splitter.py
```diff
@@ -89,6 +82,69 @@ def copy(self):
...
+                list_arrays_y.append(y_for_cluster[label][:nb_samples_to_pick])
+                y_for_cluster[label] = y_for_cluster[label][nb_samples_to_pick:]
+
+            res.append((np.concatenate(list_arrays_x), np.concatenate(list_arrays_y)))
+            nb_samples_split.append(len(np.concatenate(list_arrays_y)))
+
+        # Log the relative amounts of samples split among partners
+        total_nb_samples_split = np.sum(nb_samples_split)
+        relative_nb_samples = [round(nb / total_nb_samples_split, 2) for nb in nb_samples_split]
+        logger.info(f"Partners' relative number of samples: {relative_nb_samples}")
```
arthurPignet, line 143: Maybe more than log those, we can save it somewhere ? Maybe in a property of each partner, like datavolume, but in proportion

arthurPignet (changes requested):
What do you think of rewrite random and stratified on the top of this method ?

bowni:
> What do you think of rewrite random and stratified on the top of this method ?

Agreed! But maybe not in this PR. It could be an enhancement (or refacto let's say) afterwards.

mplc/scenario.py
bowni: That is true, it could. However, this shouldn't happen too often that someone wants to do that.

The reason I did that is that some splitters might not require those assertions, and might even be incompatible with them. That is the case of the `FlexibleSplitter` which works without any `amounts_per_partner`.

bowni: If you see a more direct and pythonic way for that, don't hesitate that would be good!

bowni: Now thinking about it, I think those asserts (or similar ones) should be **both** in `Scenario` and in `Splitter`:

- at scenario level, for a `Scenario` to be coherent, those tests need to pass otherwise the user config doesn't add up
- at splitter level, some splitters need to enforce that too, and, as you say @arthurPignet , in some cases one might want to use a splitter outside a scenario

The 2 perspectives complement each other, but both need to live independently too.

mplc/splitter.py
bowni: Yes, I'll remove that. I copied blindly on the `Advanced Splitter`, which `__str__` method is rewritten to include how it is configured. It doesn't seem that useful actually

mplc/splitter.py
bowni: In principle I would say why not. But I can't think of an actual use of this info. Moreover, it is dependent upon the `Scenario`, so putting it as a `Partner` property, I am not sure. I'll leave it like that for now

mplc/scenario.py
bowni: So I added an abstract method for coherence test at the beginning of the `split()` function. Each Splitter has to implement its own coherence tests within it (although 2 of them are generic enough to be implemented within the abstract class)

mplc/splitter.py
bowni: Yes - unless it really provides an improvement, I'll stick to the `list(zip(*...))` for now

mplc/splitter.py
```diff
@@ -37,29 +33,41 @@ def __str__(self):
...
 
         for partner in self.partners_list:
             logger.info(
-                f"   Partner #{partner.id}: "
-                f"{partner.final_nb_samples} samples "
-                f"with labels {partner.labels}"
+                f"Partner #{partner.id}: {partner.final_nb_samples} samples with labels {partner.labels}"
             )
 
+    @abstractmethod
```
arthurPignet, line 57: Technically an abstract method should never be called (even if it seems that python allows it), but must be re-defined by the subclasses. 

It is a way of saying, 'hey, you are my subclass, you should implement your own $abstractmethod$'
kind of a template.

I maybe misunderstood but I don't think that is your objective ? 

Typically all the other splitters could reuse the method implemented in the abstract class if this method wasn't abstract

mplc/doc/documentation.md
```diff
@@ -238,23 +238,25 @@ There are 2 ways to select a dataset. You can either choose a pre-implemented da
...
+  
+    - `RandomSplitter`: the dataset is shuffled and partners receive data samples selected randomly. String identifier: `'random'`
+      
+    - `StratifiedSplitter`: the dataset is stratified per class and each partner receives certain classes only (note: depending on the `amounts_per_partner` specified, there might be some overlap of classes). String identifier: `'stratified'`
+    
+    - `'AdvancedSplitter'`: in certain cases it might be interesting to split the dataset among partners in a more elaborate way. For that we consider the data samples from the initial dataset as split in clusters per data labels. The advanced split is configured by indicating, for each partner in sequence, the following 2 elements: `[[nb of clusters (int), 'shared' or 'specific']]`. Practically, you can either instantiate your `AdvancedSplitter` object, and pass this list `[[nb of clusters (int), 'shared' or 'specific']]` to the keyword argument `description`, or use the string identifier and pass the list `[[nb of clusters (int), 'shared' or 'specific']]` to the `Scenario` via the keyword argument `samples_split_configuration`. String identifier:`'advanced'`. Configuration:
```
RomainGoussault, line 248: Why do we have quotes here: 'AdvancedSplitter' and not quote in RandomSplitter?

mplc/doc/documentation.md
```diff
@@ -238,23 +238,25 @@ There are 2 ways to select a dataset. You can either choose a pre-implemented da
...
-
-  Example: `samples_split_option='advanced', samples_split_configuration=[[7, 'shared'], [6, 'shared'], [2, 'specific'], [1, 'specific']]]`
+          - `'specific'`: each partner with option `'specific'` receives data samples picked 
+                         from cluster(s) it is the only one to receive from  
+    Example: `samples_split_option='advanced', samples_split_configuration=[[7, 'shared'], [6, 'shared'], [2, 'specific'], [1, 'specific']]]`
+    
+    - `'FlexibleSplitter'`: in other cases one might want to specify in detail the split among partners (partner per partner and class per class). For that the `FlexibleSplitter` can be used. It is configured by indicating, for each partner in sequence, a list of the percentage of samples for each class: `[[% for class 1, ..., % for class n]]`. As above, it can be instantiated separately and then passed to the `Scenario` instance. Or the string identifier `'flexible'` can be used for the parameter `samples_split_option`, coupled with the split configuration passed to the keyword argument `samples_split_configuration`. String identified: `'flexible'`.
```
RomainGoussault, line 257: Same here, why the quotes?

RomainGoussault (changes requested):
Nice work! Could you add some unit tests to test the new split options?
Code coverage should not go down ;)

mplc/doc/documentation.md
bowni: Right - now removed

mplc/doc/documentation.md
bowni: Right - now removed

mplc/splitter.py
bowni: I am bit torn apart here - initially I wanted indeed to force all subclass to reimplement it, as these coherence tests are very important and could be quite different depending on the splitter. But it turned out 3 out of the current 4 use the same... I think I'll relax it, and we'll see in the future if it is best to re-make it an abstract method.

tests/unit_tests.py
```diff
@@ -113,13 +118,15 @@ def create_Partner(create_all_datasets):
                          ['not-corrupted'] * 3),
                         (Cifar10, "random", ['not-corrupted'] * 3),
                         (Cifar10,
-                         AdvancedSplitter([0.3, 0.5, 0.2], [[4, "specific"], [6, "shared"], [4, "shared"]]),
+                         FlexibleSplitter([0.3, 0.5, 0.2], [[0.33, 0.33, 0.33, 1.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0],
```
arthurPignet, line 121: 👍

arthurPignet (approved):
I will have only one question: can a flexible splitter be used with a config file ? Did you test it ?

bowni:
@RomainGoussault could you check if your "requested changes" are now OK?

natct10 (approved):
Nice work here @bowni!

RomainGoussault (approved):
looks good to me!
And we have docs AND tests for this PR 🎉 !

PR #326, opened 2021-01-26
Pytest 6.2.2

natct10:
Pytest changelog: https://docs.pytest.org/en/stable/changelog.html

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,7 +10,7 @@ seaborn==0.11.0
 loguru==0.5.3
 tensorflow==2.4.0
 ruamel.yaml==0.16.12
-pytest==6.2.0
+pytest==6.2.2
 pytest-cov==2.10.1
 flake8==3.8.3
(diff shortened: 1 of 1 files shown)


arthurPignet:
Why do you rename it requirements instead of dev-req ? There already is a requirement.txt, in the mplc folder.

arthurPignet (commented):
Ok for the pytest update, but I don't understand why renaming dev-req. There already exist a mplc/requirement.txt, which is enough to run mplc and dev-req is for running the test, flake it, and make a pip release

natct10:
> Ok for the pytest update, but I don't understand why renaming dev-req. There already exist a mplc/requirement.txt, which is enough to run mplc and dev-req is for running the test, flake it, and make a pip release

Oh my bad, I got confused, I'll revert this part

arthurPignet (approved):

arthurPignet:
👍

bowni (approved):

PR #319, opened 2021-01-12
[wip] Add Makefile

natct10:
You can use `make black`, `make flake`, `make tests` or even `make all` to run those tasks locally before pushing!
Feel free to suggest any improvement. This should work out-of-the-box on linux or mac, but windows users will have to install a build program (easy with chocolatey package manager: https://chocolatey.org/packages/make).

Close #312

PR #333, opened 2021-04-01
Makefile

### LINKED ISSUE #312: Add make build

natct10:
You can use `make black`, `make flake`, `make tests`, `make tests-ml-perf` and `make test-contrib`. Please install `make` to test. 
Windows users might use chocolatey package manager: https://chocolatey.org/packages/make.

Started here: https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/319
Fix #312

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+black:
+	black .
+
+flake:
+	flake8 --max-line-length=120
+
+tests:
+	pytest -vv tests/ml_perf_end_to_end_tests.py
+	pytest -vv tests/contrib_end_to_end_tests.py
+
--- a/Makefile
+++ b/Makefile
@@ -8,3 +8,5 @@ tests:
 	pytest -vv tests/ml_perf_end_to_end_tests.py
 	pytest -vv tests/contrib_end_to_end_tests.py
 
+all: black flake tests
+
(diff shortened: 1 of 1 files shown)


bowni (changes requested):
Cool thing in principle. But I would have a couple remarks:

1. Could we remove the black formatter? I really don't like the syntax it gives, it makes code files difficult to read

1. As discussed this morning, I am not too sure about end-to-end tests on this one, it might take a while, and become useless because of that. I think flake + unit tests would be a good combo, both handy and useful

natct10:
> Cool thing in principle. But I would have a couple remarks:
> 
>     1. Could we remove the black formatter? I really don't like the syntax it gives, it makes code files difficult to read

Black is a great tool. And you don't have to accept all, `git add -p` is useful in this regard.

>     2. As discussed this morning, I am not too sure about end-to-end tests on this one, it might take a while, and become useless because of that. I think flake + unit tests would be a good combo, both handy and useful

This part is refined in this PR: https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/336/files. The idea was -even if the command takes a long time to be executed- to have an alias to run it.

Let me know

bowni (approved):
LGTM 👍 @natct10

natct10:
Closing as it is included in [336](https://github.com/SubstraFoundation/distributed-learning-contributivity/pull/336/files)

PR #334, opened 2021-04-01
Fix typo in config_quick_debug

celinejacques:
I just noticed a mistake in config_quick_debug. There is an extra ' ] '


natct10 (approved):
Oh good catch!

bowni (approved):

PR #336, opened 2021-04-01
Tox integration as test runner

SaboniAmine:
Tox will help :
- Isolate tests, and select perimeter to launch for different contexts
- Test over multiple environnements to help integration

Main changes : 
- requirements.txt not needed anymore in the setup (are they used by another part of the application?)
- make file integration (based on #333 )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,6 @@ tests-ml_perf:
 	tox -e ml_perf
 
 tests-contrib:
-	tox -e contrib --parallel 8
+	tox -e contrib --parallel 4
 
 all: flake tests
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,12 @@ flake:
 	flake8 --max-line-length=120
 
 tests-unit:
-	tox -e unit --parallel 2
+	tox -e unit --parallel 2 --verbose
 
 tests-ml_perf:
-	tox -e ml_perf
+	tox -e ml_perf --verbose
 
 tests-contrib:
-	tox -e contrib --parallel 2
+	tox -e contrib --parallel 2 --verbose
 
-all: flake tests
+all: flake tests-unit tests-ml_perf tests-contrib
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+flake:
+	flake8 --max-line-length=120
+
+tests-unit:
+	tox -e unit --parallel 8
+tests-ml_perf:
+	tox -e ml_perf
+
+tests-contrib:
+	tox -e contrib --parallel 8
+
+all: flake tests
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,13 @@
-#!/usr/bin/env bash
 flake:
 	flake8 --max-line-length=120
 
 tests-unit:
-	tox -e unit --parallel 4
+	tox -e unit --parallel 2
 
 tests-ml_perf:
 	tox -e ml_perf
 
 tests-contrib:
-	tox -e contrib --parallel 4
+	tox -e contrib --parallel 2
 
 all: flake tests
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -3,8 +3,30 @@
...
+    'scikit-learn==0.23.2',
+    'pandas==1.1.5',
+    'seaborn==0.11.0',
+    'loguru==0.5.3',
+    'tensorflow==2.4.1',
+    'ruamel.yaml==0.16.12',
+    'flake8==3.8.3',
+    'setuptools==50.3.0',
+    'wheel==0.35.1',
+    'twine==3.2.0'
```
natct10, line 22: twine is only for the publication of the package on pypi servers, I wouldn't add it here, no?

setup.py
```diff
@@ -3,8 +3,30 @@
...
+]
+
+test_requirements = [
+    'tox',
```
natct10, line 26: It would be nice to have a version for tox, could you please add one?

setup.py
SaboniAmine: Should I extract all dev/packaging dependencies in a custom install ?

natct10: I think that having the dev-requirements.txt in a specific file is fine as it is only used for packaging and publishing new builds

tox.ini
```diff
@@ -0,0 +1,41 @@
...
+[testenv:contrib]
+
+deps =
+    pytest
+
+recreate = false
+
+commands =
+    pip install -e .
+    pytest -vv tests/contrib_end_to_end_test.py
```
natct10, line 24: I think the coverage part (`--cov=mplc` & `--cov=mplc --cov-append`) is missing here and in line 36 or it should be added only to the CI yaml!

natct10 (changes requested):
This looks cool, thank you @SaboniAmine! I will test locally!

bowni (approved):
LGTM, 👍 @SaboniAmine

bowni:
Someone else to review @natct10 ? @RomainGoussault ? @celinejacques ?

setup.py
```diff
@@ -3,8 +3,28 @@
...
+    'scipy==1.5.4',
+    'scikit-learn==0.23.2',
+    'pandas==1.1.5',
+    'seaborn==0.11.0',
+    'loguru==0.5.3',
+    'tensorflow==2.4.1',
+    'ruamel.yaml==0.16.12',
+    'flake8==3.8.3',
+    'setuptools==50.3.0',
+    'wheel==0.35.1']
```
natct10, line 21: I think `flake8`, `setuptools` and `wheel` are required respectively for linting and building, I am not sure that it needs to be included in the `setup.py`, but I am not sure about this

Makefile
```diff
@@ -0,0 +1,13 @@
+flake:
+	flake8 --max-line-length=120
+
+tests-unit:
```
natct10, line 4: Is there a way to make it verbose and to see the name of tests running?

natct10: Yes, in fact, it is possible to use `tox -e unit -- --debug`, do you think it would be relevant to add another make command like `tests-unit-verbose` (with the same suffix for others),

natct10: Oh and there is also `--verbose`!

setup.py
SaboniAmine: I agree, as they are in dev-requirements they can be removed here. Thanks for noticing it.

Makefile
SaboniAmine: I suggest to let the verbose option by default, as it doesn't seem to flood that much.

natct10: Okay cool, thank you! Would you kindly add the `tox -e unit -- --debug` command to the readme, so we have the command for running tests with tox with the full output?

natct10: I mean to the contributing doc :D

natct10 (approved):
That's nice, thank you @SaboniAmine!

Issue #331, opened 2021-03-02
Clarify in README and/or documentation when results are saved


bowni:
And/or also add saving results of scenarios run alone

PR #337, opened 2021-04-01
Change results saving behaviour with results saved by default

### LINKED ISSUE #329: Clarify in documentation the use of save_path in Scenario and Experiment
As saving results is not by default (see in code `:param save_path: path where to save the scenario outputs. By default, they are not saved!`), we should make it very clear and explicit in the documentation.

bowni:
Changes:

- Results are now always saved
- Two cases taken into account: whether a scenario is run as standalone or as part of an Experiment

Fixes #329, #331


JustineBoulant (approved):
+1

mplc/constants.py
```diff
@@ -19,7 +19,8 @@
 
 # Paths
 EXPERIMENTS_FOLDER_NAME = "experiments"
-EXPERIMENT_RESULT_FILE_NAME = "results.csv"
+SINGLE_SCENARIOS_FOLDER_NAME = "standalone_scenarios"
+RUN_RESULT_FILE_NAME = "results.csv"
```
RomainGoussault, line 23: Would rename to `RESULTS_FILE_NAME`

mplc/scenario.py
```diff
@@ -393,14 +412,12 @@ def copy(self, **kwargs):
...
+                    'splitter',
+                    ]:
             del params[key]
         if 'is_quick_demo' in kwargs and kwargs['is_quick_demo'] != self.is_quick_demo:
             raise ValueError("Attribute 'is_quick_demo' cannot be modified between copies.")
-        if self.save_folder is not None:
-            params['save_path'] = self.save_folder.parents[0]
-        else:
-            params['save_path'] = None
+        params['save_path'] = self.save_folder.parents[0]
```
RomainGoussault, line 420: `self.save_folder.parent` should do the same think and be a bit more concise.

RomainGoussault (commented):
Could we add some test for this @bowni?
I was thinking of running an experiment and checking there is a result file that was created. (and same thing for running a scenario).

mplc/constants.py
bowni: OK

mplc/scenario.py
bowni: It was done already in one of the PRs recently merged - got it with the rebase.

JustineBoulant (approved):

RomainGoussault:
> Could we add some test for this @bowni?
> I was thinking of running an experiment and checking there is a result file that was created. (and same thing for running a scenario).

@bowni I think it's still useful to add a test. What do you think?

RomainGoussault (approved):
LGTM !

PR #338, opened 2021-04-02
Testing nightly builds

SaboniAmine:
As far as I understood, scheduled build can only be activated on the workflow on the default branch, master.
That means that it cannot be executed from a non-default branch, thus quite hard to reproduce (even locally, I tried with [act](https://github.com/nektos/act)).
I propose to merge this lone update, after changing the schedule to a convenient hour, to test it live.

Another PR could be open to segregate heavy tests only for the nightly builds.

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -3,7 +3,7 @@ name: Build
 on:
   push:
   schedule:
-    - cron: '0 0 * * * '
+    - cron: '0 0 * * * ' # Every day at 00.00
 
 jobs:
   build:
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -1,6 +1,9 @@
 name: Build
 
-on: [push]
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * * '
 
 jobs:
   build:


SaboniAmine:
Finally could successfully test the scheduling with act. Here is the setup to reproduce :
- Install act via brew or [install script](https://github.com/nektos/act/blob/master/install.sh)
- Patch the default ubuntu image, which [cannot handle](https://github.com/nektos/act/issues/418#issuecomment-727261137) python versions matrix iterations properly, with the following dockerfile : 

```dockerfile
FROM ubuntu:20.04

RUN apt-get update \
    && apt update -y \
    && apt-get upgrade -y \
    && DEBIAN_FRONTEND=noninteractive apt-get install -y \
    build-essential \
    curl \
    nodejs \
    npm \
    sudo \
    libsndfile1 \

    && rm -rf /var/lib/apt/lists/*
```
- Build image as ubuntu-builder : 
```bash
docker build . -t local/ubuntu-builder:latest
```
- Run locally workflow main_ci.yml with patched image as CI plateform (from project root) : 
```bash
act -v schedule -P ubuntu-latest=local/ubuntu-builder:latest 
```

Test should run 3 parallel builds at scheduled time, and 3 parallel tensorflow downloads, which could be bandwith / CPU intensive.

bowni (approved):
Well done @SaboniAmine, thank you 👍 
LGTM but I think a review from @RomainGoussault , with a deeper look on consequences on the CI execution, is needed!

.github/workflows/main_ci.yml
```diff
@@ -1,6 +1,9 @@
 name: Build
 
-on: [push]
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * * '
```
natct10, line 6:
```suggestion
    - cron: '0 0 * * * ' # Every day at 00.00
```

natct10 (commented):
I would just add a comment for the cron job.

natct10 (approved):
Let's try this!

PR #345, opened 2021-05-22
Federated averaging with local and global optimizations

arthurPignet:
This PR follows the draft 235, which was too far behind master. This is a rebased version.  

# New mpl methods:

FedGDO stands for Federated Gradient Double Optimization.

This method is inspired from Federated gradient, but with modification on the local computation of the gradient.
In this version we use a local optimizer (partner-specific) to do several minimization steps of the local-loss
during a minibatch. We use the sum of these weighs-updates as the gradient which is sent to the global optimizer.
The global optimizer aggregates these gradients-like which have been sent by the partners,
and performs a optimization step with this aggregated gradient.
Here three variations of this mpl method are tested.
 - FedGDO with fresh local optimizer at each minibatch
 - FedGDO with persistent local optimizer

The reset of the optimizers can be set via a parameter of the mpl methods. In the same way, a global optimizer different from the local ones can be passed via the mpl arguments. 

These methods are tested on this notebook -> https://colab.research.google.com/drive/1CcQpWRpLGldj3iNR7v7Hv2brdwBP3z7D?usp=sharing

Please note that as I am currently working with the notebook, it can change, and be not fully readable 
Notebook access is currently limited to substra.org, but don't hesitate to come to me for access.

### DIFF
--- a/mplc/multi_partner_learning/fast_mpl.py
+++ b/mplc/multi_partner_learning/fast_mpl.py
@@ -588,9 +588,9 @@ class FastFedGDO(FastFedAvg):
     name = 'FastFedGDO'
 
     def __init__(self, scenario, reset_local_optims=False, global_optimiser=None, **kwargs):
-        super(FastFedGDO, self).__init__(scenario, **kwargs)
         self.global_optimiser = global_optimiser
         self.reset_local_optims = reset_local_optims
+        super(FastFedGDO, self).__init__(scenario, **kwargs)
 
     def init_specific_tf_variable(self):
         # generate tf Variables in which we will store the model weights
--- a/mplc/multi_partner_learning/fast_mpl.py
+++ b/mplc/multi_partner_learning/fast_mpl.py
@@ -602,6 +602,8 @@ class FastFedGDO(FastFedAvg):
         self.partners_optimizers = [self.model.optimizer.from_config(self.model.optimizer.get_config()) for _ in
                                     self.partners_list]
         if self.global_optimiser:
+            # If the global optimizer is specified in the kwargs. If not the global optimizer used is the
+            # default one, same as the local optimizer, and we don't need to re-compile the model
             self.model.compile(optimizer=self.global_optimiser)
 
     def fit(self):
--- a/mplc/multi_partner_learning/fast_mpl.py
+++ b/mplc/multi_partner_learning/fast_mpl.py
@@ -575,3 +575,92 @@ class FastGradSmodel(FastFedGrad):
                 break
 
         self.log_end_training()
+
+
+class FastFedGDO(FastFedAvg):
+    """
+     This method is inspired from Federated gradient, but with modification on the local computation of the gradient.
+    In this version we use a local optimizer (partner-specific) to do several minimization steps of the local-loss
+    during a minibatch. We use the sum of these weighs-updates as the gradient which is sent to the global optimizer.
+    The global optimizer aggregates these gradients-like which have been sent by the partners,
+    and performs a optimization step with this aggregated gradient.
+    """
+    name = 'FastFedGDO'
+
+    def __init__(self, scenario, reset_local_optims=False, **kwargs):
+        super(FastFedGDO, self).__init__(scenario, **kwargs)
+        self.reset_local_optims = reset_local_optims
+
+    def init_specific_tf_variable(self):
+        # generate tf Variables in which we will store the model weights
+        self.model_stateholder = [tf.Variable(initial_value=w.read_value()) for w in self.model.trainable_weights]
+        self.partners_grads = [[tf.Variable(initial_value=w.read_value()) for w in self.model.trainable_weights]
+                               for _ in self.partners_list]
+        self.global_grad = [tf.Variable(initial_value=w.read_value()) for w in self.model.trainable_weights]
+        self.partners_optimizers = [self.model.optimizer.from_config(self.model.optimizer.get_config()) for _ in
+                                    self.partners_list]
+
+    def fit(self):
+        # TF function definition
+        @tf.function
+        def fit_minibatch(model, model_stateholder, partners_minibatches, partners_optimizers, partners_grads,
+                          global_grad, aggregation_weights):
+            for model_w, old_w in zip(model.trainable_weights, model_stateholder):  # store model weights
+                old_w.assign(model_w.read_value())
+
+            for p_id, minibatch in enumerate(partners_minibatches):  # minibatch == (x,y)
+                # minibatch[0] in a tensor of shape=(number of batch, batch size, img).
+                # We cannot iterate on tensors, so we convert this tensor to a list of
+                # *number of batch* tensors with shape=(batch size, img)
+                x_minibatch = tf.unstack(minibatch[0], axis=0)
+                y_minibatch = tf.unstack(minibatch[1], axis=0)  # same here, with labels
+
+                for x, y in zip(x_minibatch, y_minibatch):  # iterate over batches
+                    with tf.GradientTape() as tape:
+                        y_pred = model(x)
+                        loss = model.compiled_loss(y, y_pred)
+                    model.compiled_metrics.update_state(y, y_pred)  # log the loss and accuracy
+                    partners_optimizers[p_id].minimize(loss, model.trainable_weights,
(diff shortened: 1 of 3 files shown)


bowni (approved):
OK, well done @arthurPignet this looks very promising (although not always easy to picture and understand line by line) 💪 
@RomainGoussault @jeromechambost @celinejacques @SaboniAmine @HeytemBou @JustineBoulant @kat-leen who would like to do another review please?

bowni:
@RomainGoussault proposes to review it!

mplc/multi_partner_learning/fast_mpl.py
```diff
@@ -575,3 +575,95 @@ def fit_epoch(model, train_dataset, partners_grads, smodel_list, global_grad, ag
                 break
 
         self.log_end_training()
+
+
+class FastFedGDO(FastFedAvg):
+    """
+     This method is inspired from Federated gradient, but with modification on the local computation of the gradient.
```
RomainGoussault, line 582: I would add that FedGDO stand for Federated averaging with double optimizers.

mplc/multi_partner_learning/fast_mpl.py
```diff
@@ -575,3 +575,95 @@ def fit_epoch(model, train_dataset, partners_grads, smodel_list, global_grad, ag
...
+
+    def __init__(self, scenario, reset_local_optims=False, global_optimiser=None, **kwargs):
+        self.global_optimiser = global_optimiser
+        self.reset_local_optims = reset_local_optims
+        super(FastFedGDO, self).__init__(scenario, **kwargs)
+
+    def init_specific_tf_variable(self):
+        # generate tf Variables in which we will store the model weights
+        self.model_stateholder = [tf.Variable(initial_value=w.read_value()) for w in self.model.trainable_weights]
+        self.partners_grads = [[tf.Variable(initial_value=w.read_value()) for w in self.model.trainable_weights]
```
RomainGoussault, line 598: this is gradients or weights?

mplc/multi_partner_learning/fast_mpl.py
```diff
@@ -575,3 +575,95 @@ def fit_epoch(model, train_dataset, partners_grads, smodel_list, global_grad, ag
...
+                               for _ in self.partners_list]
+        self.global_grad = [tf.Variable(initial_value=w.read_value()) for w in self.model.trainable_weights]
+        self.partners_optimizers = [self.model.optimizer.from_config(self.model.optimizer.get_config()) for _ in
+                                    self.partners_list]
+        if self.global_optimiser:
```
RomainGoussault, line 603: in which case we won't have a global optimizer?

mplc/multi_partner_learning/fast_mpl.py
```diff
@@ -575,3 +575,95 @@ def fit_epoch(model, train_dataset, partners_grads, smodel_list, global_grad, ag
...
+            # at the end of the minibatch, aggregate all the local grads
+            for i, grads_per_layer in enumerate(zip(*partners_grads)):
+                global_grad[i].assign(tf.tensordot(grads_per_layer, aggregation_weights, [0, 0]))
+
+            # perform one optimization update using the aggregated gradient
+            model.optimizer.apply_gradients(
+                zip(global_grad, model.trainable_weights))
+
+            # Execution
+
```
RomainGoussault, line 646: blanck line to remove?

RomainGoussault (changes requested):

bowni:
@arthurPignet let us know when you can tackle @RomainGoussault 's questions and comments! 😃

mplc/multi_partner_learning/fast_mpl.py
arthurPignet: The first one is for the weights, to store initial weight of the global network at the beginning of the global batch while one partner is optimizing the network within its local batch
The second one is for gradients 

But I use the weights for both because it is convenient as the size of gradients is the same as the size of weights.

mplc/multi_partner_learning/fast_mpl.py
arthurPignet: 👍

mplc/multi_partner_learning/fast_mpl.py
arthurPignet: When the global optimizer is not specified in the kwargs. In that case the global optimizer used is the default one, same as the local optimizer, and we don't need to re-compile the model

mplc/multi_partner_learning/fast_mpl.py
arthurPignet: 👍

mplc/multi_partner_learning/fast_mpl.py
```diff
@@ -575,3 +575,95 @@ def fit_epoch(model, train_dataset, partners_grads, smodel_list, global_grad, ag
...
+                    with tf.GradientTape() as tape:
+                        y_pred = model(x)
+                        loss = model.compiled_loss(y, y_pred)
+                    model.compiled_metrics.update_state(y, y_pred)  # log the loss and accuracy
+                    partners_optimizers[p_id].minimize(loss, model.trainable_weights,
+                                                       tape=tape)  # perform local optimizations
+                # get the gradient as theta_before_minibatch - theta_after_minibatch
+                for grad_per_layer, w_old, w_new in zip(partners_grads[p_id], model_stateholder,
+                                                        model.trainable_weights):
+                    grad_per_layer.assign((w_old - w_new))
```
HeytemBou, line 632: @arthurPignet Here you're using weight differences between two consecutive minibatch updates as gradients, right? but how does that work on a theoretical level?

arthurPignet: Yes indeed. 
When you are performing a classic SGD update, the difference between the weights is the stocastic estimation of the gradient * lr. 
But here we are using more sophisticated optimization algorithms, with momentums and adaptative lr. The gradient is not easy to reach. So I fall back to an object which is not exactly the gradient, but the cumulative descent direction (which is the object of our interest actually). I compute the descent direction to be sent to the server as the vec. sum of the descents applied during the local batches. That corresponds to the difference between the weights at the beginning of the local batches and the ones at the ends of the local batches, just before the aggregation/global optimization which will conclude the global batch.  

Did I answer your question ? Maybe I misunderstood what you were saying.

HeytemBou: Yes thanks, it's easy to miss that the difference between the weights equates to gradient*lr, unless you have the SGD update rule in front of your eyes :). parhaps it would be much clearer for future readers to  to change the name of the parameter "global_grad" to something like cumulative_descent maybe, just like you explained. It's a bit misleading at first when reading the name 'grad' being used.

arthurPignet: Ok that is a nice idea. 
But I will use this cumulative_descent like a gradient just after for the aggregation and the global optimization. Don't you think that using a object which is not named grad where it is usually a gradient in every other algorithm, to play the role of the gradient could be even more misleading ?
Maybe rather than a reformating the variable I could add a comment saying that this thing is not really a gradient but a cumulative descent ?

HeytemBou: yes, a comment would be great.

bowni: Or name it `pseudo_gradient` or something of that sort

bowni:
@HeytemBou beyond your above comment and the associated thread, was your review conclusive? Are you confortable with validating the PR or not?

HeytemBou:
> @HeytemBou beyond your above comment and the associated thread, was your review conclusive? Are you confortable with validating the PR or not?

Yes

HeytemBou (approved):

arthurPignet:
> @arthurPignet let us know when you can tackle @RomainGoussault 's questions and comments! 😃

I took into account @RomainGoussault's comments, it's ready for a final (well I hope it will be) review !

mplc/multi_partner_learning/fast_mpl.py
RomainGoussault: Maybe we can add a small comment to say what you just wrote?

RomainGoussault (approved):
LGTM8

Issue #351, opened 2021-07-05
Unable to run single partner learning approach


HeytemBou:
https://github.com/SubstraFoundation/distributed-learning-contributivity/blob/4fb247a35ae40ce27790b888cc6e4cb52134fa68/mplc/multi_partner_learning/basic_mpl.py#L350

@arthurPignet Is this comment still valid or is it no longer possible to use the Single Partner Learning approach when running a scenario?

arthurPignet:
I just checked, the mpl method single partner learning does exist, (in basic mpl) but it is not reachable. I will fix it. 
However this comment is not really relevant anymore, the method raises an error if only one partner is provided, and doesn't fall back to the dedicated single partner class

PR #352, opened 2021-07-05
Add single partner learning to available methods

arthurPignet:
Fix issue #351 


bowni (approved):

HeytemBou (approved):

HeytemBou (approved):
LGTM!

tests/ml_perf_end_to_end_tests.py
```diff
@@ -57,13 +57,21 @@ def test_all_mpl_approaches(self):
 
         exp = Experiment(is_save=False)
         # mpl_approaches = mpl.BASIC_MPL_APPROACHES.copy()
-        mpl_approaches = mpl.MULTI_PARTNER_LEARNING_APPROACHES
-
+        mpl_approaches = mpl.MULTI_PARTNER_LEARNING_APPROACHES.copy()
+        _ = mpl_approaches.pop("single-partner")
```
RomainGoussault, line 61: I understand we can not test this method with the example below with 2 partners but we should nonetheless test it. I guess we would have to do a dedicated test with one partner.

RomainGoussault (changes requested):

tests/ml_perf_end_to_end_tests.py
bowni: Good catch @RomainGoussault , I missed it but I agree with you. @arthurPignet is this something you could add before we merge?

arthurPignet: In fact it is tested within this function. See lines 66-73. I just cannot use the loop, because I want only one partner.

PR #353, opened 2021-07-15
Plot data distribution

kat-leen:
New plot of the data/ground truth distribution based on the one made by @bowni for the reference scenarios. 

Changes :
- The figure was entirely rebuilt with seaborn.
- We can visualize both the real ground truth distribution and the corrupted one if there is any corrupted partner.
- We can now display the plot after the scenario creation.
- The plots were added in the notebook of the reference scenarios.
- The code is divided into two parts, one in the Scenario class and the other one (more independent) in utils. But is it really relevant? I'm not sure whether it can be reused for something else or not.
- The documentation was updated.


mplc/doc/documentation.md
```diff
@@ -109,6 +110,16 @@ my_scenario = Scenario(partners_count=3,
...
+### Plot data distribution
+
+If you whish to visualize the partners' ground truth distribution of the train set per class (with and without corruption if there is any), you can use the method `plot_data_distribution()`. You may choose whether you want to display the plot straight away (`display` is `True` by default) or save it (`save` is `False` by default).
+
+```python
+my_scenario.plot_data_distribution(save=False, display=True)
+```
+
+> Note: the plot is automatically saved when calling the `run()` method described below.
+
```
bowni, line 122: Very clear - seems good to me 👍

bowni, line 122: To provide an example to the reader, could you @kat-leen add an example graph here and/or add a link to the notebook?

mplc/utils.py
```diff
@@ -12,8 +12,19 @@
...
+import pandas as pd
+import matplotlib.pyplot as plt
+import seaborn as sns
+import pylab
+
 from . import constants
 
+sns.set_style("white")
+pylab.rcParams.update({'font.size': 18})
+
```
bowni, line 27: Could you add a comment/title to these 2 lines so that they don't sit here isolated? 😃

mplc/utils.py
```diff
@@ -156,3 +167,88 @@ def set_log_file(path):
...
+               .assign(after_part=lambda df: 1 - df['partner_part'].cumsum())\
+               .reindex(['before_part', 'partner_part', 'after_part'], axis=1)\
+               .cumsum(axis=1).add(i, axis=1)\
+               .rename_axis('partner_num').reset_index()\
+               .melt(id_vars=['partner_num'], value_vars=['partner_part', 'before_part', 'after_part'],
+                     var_name='chunk', value_name='value')
+        for part, cols in zip(["after_part", "partner_part", "before_part"],
+                              [['white'] * num_partners, palette, ['white'] * num_partners]):
+            bars = sns.barplot(data=df.query('chunk=="' + part + '"'), y='partner_num', x='value',
+                               orient='h', palette=cols, ci=None, edgecolor=hatch_color, linewidth=0, )
```
bowni, line 211: That is quite impressive!

mplc/utils.py
```diff
@@ -156,3 +167,88 @@ def set_log_file(path):
...
+
+    # Add corruption hatches
+    corr_labels = [''] * num_partners
+    for corr, ratio in corrupted_partners.items():
+        selection = [c * num_partners * 3 + 1 * num_partners + corr for c in range(num_classes)]
+        keys = np.array(list(hatches.keys()))
+        hatch = hatches[keys[np.argmin(np.abs(keys - ratio))]]
+        for s in selection:
+            bars.patches[s].set_hatch(hatch)
+        corr_labels[corr] += f'\n({int(ratio * 100)}% corrupted data)'
```
bowni, line 221: I am not experienced enough with Seaborn to understand this piece clearly - but it seems to me that dots representing corrupted data are applied to the whole bar of each class for which a partner has its data corrupted, right?

mplc/doc/documentation.md
```diff
@@ -109,6 +110,16 @@ my_scenario = Scenario(partners_count=3,
                        minibatch_count=3)
 ```
 
+### Plot data distribution
+
+If you whish to visualize the partners' ground truth distribution of the train set per class (with and without corruption if there is any), you can use the method `plot_data_distribution()`. You may choose whether you want to display the plot straight away (`display` is `True` by default) or save it (`save` is `False` by default).
```
HeytemBou, line 115: typo error "whish"

HeytemBou (approved):

bowni (approved):

bowni:
Well done @kat-leen great addition to the library for much clearer definition and execution of scenarios 👍

